PR #9, opened 2019-10-20
fix parse error, removed line in header.html

boris779:
As already mentioned, I get a parser error and cannot start hugo server.

`
boris@xxxx:~/repos/selenium/site/docs_source_files$ hugo server
Error: "xxxxx/repos/selenium/site/docs_source_files/themes/hugo-theme-learn/layouts/partials/header.html:6:1": parse failed: template: partials/header.html:6: function "hugo" not defined
`

When I remove the line, everything is fine on my machine (ubuntu)


boris779:
Hi. I did a mistake. My PR should only contain one commit. Sorry. I will do it again.

Issue #19, opened 2019-10-26
Add new language on site (Korean)


0420syj:
Hello. I am a university student in Korea. First of all, thank you for your wonderful project and documentation.
I've recently learned about crawler including this for team project, and my friends and I were really impressed by your documentation. And I would like to add Korean to your document. Because there'are lot of people who are interested in crawler this could help more Korean developer to use Selenium.
I am raising an issue because I want to hear your opinion on this. Thank you for reading the long post.

PS. And I have also signed the CLA on this [link](https://github.com/SeleniumHQ/docs/blob/gh-pages/README.md).

diemol:
Hi @0420syj @lionking6792

Thank you for offering to translate our documentation to Korean 🙌 🎉 

Have a look at the contributing guidelines [here](https://selenium.dev/documentation/en/contributing/). If you need any help, please let us know.

lionking6792:
Hi! @diemol 

Thanks for your comment and advice.
I'll PR Korean version of docs as soon as possible!

Thanks :)

diemol:
Fixed via #68, thanks @0420syj

PR #65, opened 2019-11-06
Add Korean Directory on Documentation

lionking6792:
Add ko (Korean Directory that is copy of en directory ) on Documentation and I'll translate most of all in Korean within 3 weeks.


lionking6792:
I want to translate Korean version of documents. But, this is too big size, so I temporary made ko directory..

Is it anything I missing that requires PR?

diemol:
Hey @lionking6792,
Thanks for this PR, but sadly those are not the files that need to be changed for translation. That is why I mentioned in #19 to wait a couple of days so we can properly add instructions. 

The files need to be placed here, https://github.com/SeleniumHQ/seleniumhq.github.io/tree/master/docs_source_files/content. The docs are a Hugo site, and if you go through the directory structure, probably you can see how they should be placed. 

I'll close this PR so you can open a new one with the right files, feel free to wait a couple of days to check the complete instructions.

lionking6792:
Thanks for comment!

I really don't know what to translate but, now I catch it!! Thanks!

I'll re do it as soon as possible and PR to you

Thanks a lot :)

Issue #189, opened 2020-01-27
Remove deprecated Selenium class from Documentation code examples


marksnelson09:
## 🐛 Bug Report
The Selenium class is used in the Documentation code example on the 
'Guidelines > Page object models' page and possibly elsewhere. This was confusing, since I don't recall it being explained or referenced in the Documentation flow prior to this page. Then, I found in the API that the class is deprecated, which explained why it's usage isn't explained.

References to the Selenium class might also occur on other pages. Though, I might be confused by examples that I encountered on other websites.

## To Reproduce
https://selenium.dev/documentation/en/guidelines_and_recommendations/page_object_models/
![Selenium bug](https://user-images.githubusercontent.com/60354885/73191819-bd905100-40ed-11ea-8f77-5e46c0abfb3d.png)

## Expected behavior
Please update the code examples to not reference deprecated classes. Thanks

## Environment
* OS:  Windows
* Browser: Edge
* Browser version: 44.18362.449.0

diemol:
Thanks for raising this @marksnelson09, would you like to give us a hand and send a PR fixing the code samples?

marksnelson09:
I'm brand new to Selenium, and this is my first time using with GitHub. Perhaps, if I grow comfortable with my understanding of Selenium, I can come back around to figure out what the code examples should be.

diemol:
@marksnelson09 why don't you have a look to https://selenium.dev/documentation/en/contributing/? There are clear instructions on how to get started and send us contributions.

PR #221, opened 2020-02-12
Remove deprecated Selenium class and use Webdriver class

kzhirata:
### Description

fix https://github.com/SeleniumHQ/seleniumhq.github.io/issues/189 .
[Page object models](https://selenium.dev/documentation/en/guidelines_and_recommendations/page_object_models/) page uses deprecated classes such as the Selenium class.  
So, I fix the code samples using Webdriver Interface.

### Types of changes

- [ ] Improved translation

### Checklist


docs_source_files/content/guidelines_and_recommendations/page_object_models.de.md
```diff
@@ -65,65 +67,94 @@ Applying the page object techniques, this example could be rewritten like this
...
-                selenium.click("sign-in");
-                selenium.waitForPageToLoad("waitPeriod");
-
-                return new HomePage(selenium);
-        }
+  protected static WebDriver driver;
+
+  public SignInPage(WebDriver driver){
+    this.driver = driver;
+    PageFactory.initElements(driver, this);
```
diemol: Could you please remove the use of `PageFactory`? A normal approach declaring locators and using them in methods should be ok.
At the Selenium project, we are encouraging people to stop using `PageFactory` since it creates sometimes several `StaleElementException`, which is kind of hard to debug when you are starting with Selenium.

kzhirata: @diemol Thank you for the review! 
I've corrected what you pointed out, removing PageFactory class.

kzhirata:
@diemol Thank you for the comment. I corrected the point that was pointed out.

PR #301, opened 2020-03-11
Fixing invalid assignments in java code

praveen4463:
Color loginButtonColour = driver.findElement(By.id("login")).getCssValue("color");
Assignment like this is wrong, should use static function 'fromString' to get Color out of a string. Probably other language samples require fix too.

### DIFF
--- a/docs_source_files/content/support_packages/working_with_colours.en.md
+++ b/docs_source_files/content/support_packages/working_with_colours.en.md
@@ -151,22 +151,22 @@ Color loginButtonColour = Color.fromString(driver.findElement(By.id("login")).ge
 Color loginButtonBackgroundColour = Color.fromString(driver.findElement(By.id("login")).getCssValue("background-color"));
   {{< / code-panel >}}
   {{< code-panel language="python" >}}
-login_button_colour = driver.find_element_by_id('login').value_of_css_property('color')
-login_button_background_colour = driver.find_element_by_id('login').value_of_css_property('background-color');
+login_button_colour = Color.from_string(driver.find_element_by_id('login').value_of_css_property('color'))
+login_button_background_colour = Color.from_string(driver.find_element_by_id('login').value_of_css_property('background-color'))
   {{< / code-panel >}}
   {{< code-panel language="csharp" >}}
 // We don't have a C# code sample yet -  Help us out and raise a PR
   {{< / code-panel >}}
   {{< code-panel language="ruby" >}}
-login_button_colour = driver.find_element(id: 'login').css_value('color')
-login_button_background_colour = driver.find_element(id: 'login').css_value('background-color');
+login_button_colour = Color.from_string(driver.find_element(id: 'login').css_value('color'))
+login_button_background_colour = Color.from_string(driver.find_element(id: 'login').css_value('background-color'))
   {{< / code-panel >}}
   {{< code-panel language="javascript" >}}
 // We don't have a JavaScript code sample yet -  Help us out and raise a PR
   {{< / code-panel >}}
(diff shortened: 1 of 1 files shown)


praveen4463:
@diemol sure, please check.

docs_source_files/content/support_packages/working_with_colours.en.md
```diff
@@ -147,8 +147,8 @@ and converted into a valid Color object:
 
 {{< code-tab >}}
   {{< code-panel language="java" >}}
-Color loginButtonColour = driver.findElement(By.id("login")).getCssValue("color");
-Color loginButtonBackgroundColour = driver.findElement(By.id("login")).getCssValue("background-color");
+Color loginButtonColour = Color.fromString(driver.findElement(By.id("login")).getCssValue("color"));
```
Harsha509:
```suggestion
String loginButtonColour = driver.findElement(By.id("login")).getCssValue("color");
```
I have checked the code sample with selenium alpha4, This will the fix the above. I think there is no need to use `staticFunction` to get the color

Also,
String loginButtonColour = driver.findElement(By.id("login")).getCssValue("color");
(and)
Color loginButtonColour = Color.fromString(driver.findElement(By.id("login")).getCssValue("color"));

return the same output.

docs_source_files/content/support_packages/working_with_colours.en.md
```diff
@@ -151,22 +151,22 @@ Color loginButtonColour = Color.fromString(driver.findElement(By.id("login")).ge
 Color loginButtonBackgroundColour = Color.fromString(driver.findElement(By.id("login")).getCssValue("background-color"));
   {{< / code-panel >}}
   {{< code-panel language="python" >}}
-login_button_colour = driver.find_element_by_id('login').value_of_css_property('color')
-login_button_background_colour = driver.find_element_by_id('login').value_of_css_property('background-color');
+login_button_colour = Color.from_string(driver.find_element_by_id('login').value_of_css_property('color'))
+login_button_background_colour = Color.from_string(driver.find_element_by_id('login').value_of_css_property('background-color'))
```
Harsha509: Hi @praveen4463,

I think there is a bit of misunderstanding. As @diemol suggested, he is referring to other translated pages like fr.md, es.md, zh-ch.md, de.md,jo.md, ka.md, nl.md etc and not for other code samples.

For example, code changes for java should be changed in a few of the below documents too. 
https://www.selenium.dev/documentation/zh-cn/support_packages/working_with_colours/
https://www.selenium.dev/documentation/de/support_packages/working_with_colours/
https://www.selenium.dev/documentation/fr/support_packages/working_with_colours/
https://www.selenium.dev/documentation/es/support_packages/working_with_colours/

Please let me know if you need any information, we can help you out.

Thanks,
Harsha

docs_source_files/content/support_packages/working_with_colours.en.md
praveen4463: no, these aren't same. `getCssValue` gives you the computed `rgba` as string whereas using the static function returns `Color` object that can be compared to any other Color object. This is useful when you need to compare two objects derived using different mechanism, such as hash value vs rgba. You can refer to the documentation. I advised a fix because, the documentation says in the next line of getting the color `assert loginButtonBackgroundColour.equals(HOTPINK);`, this can be possible only when you've a valid Color object which `getCssValue` doesn't return. Does it makes sense?

Harsha509:
Hi @praveen4463 ,

Did you get a chance to look into the changes suggested ?

Thanks,
Harsha.

PR #328, opened 2020-03-17
Modify: Translating doc to FR and minor code changes

Harsha509:
### Description
1. Translating doc from [EN] to [FR]
2. `System.setProperty("webdriver.gecko.driver", "/path/to/geckodriver")` is not required as we target examples assuming that drivers are already set in path, also test targets in chrome and setting gecko path is wrong in example

### Types of changes

- [x] Improved translation

### Checklist


Harsha509:
> Thank you @Harsha509!

Thanks for review @diemol

PR #351, opened 2020-03-21
Add: Proposing Legacy Capabilities document

Harsha509:
### Description
Proposing Legacy Capabilities document, which speaks about deprecation of `desiredCapabilities` and `requiredCapabilities`.

We can also include using of respective [driver]:options with code samples to set capabilities instead of desiredCapabilities

### Types of changes

- [x] Improved translation

### Checklist

### DIFF
--- a/docs_source_files/content/driver_idiosyncrasies/legacy_capabilities.de.md
+++ b/docs_source_files/content/driver_idiosyncrasies/legacy_capabilities.de.md
@@ -1,0 +1,26 @@
+---
+title: "Legacy capabilities"
+weight: 1
+---
+
+{{% notice info %}}
+<i class="fas fa-language"></i> Diese Seite wird von Englisch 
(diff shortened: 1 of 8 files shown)


Harsha509:
Requesting @manoj9788  and @diemol for review!

Harsha509:
> Hi @Harsha509,
> 
> I like the intention of the PR, which I feel is to illustrate the history and how capabilities have evolved, and show how desired capabilities are being deprecated and so on...
> 
> But, if that is really the intention, this PR needs to be more complete, but I am not sure how valuable this content would be for users...
> 
> I would find more useful a set of examples with old capabilities code and examples with new capabilities. E.g. this is how it looks in old deprecated caps, and this is how it looks with options. And maybe name the section something like "from desired capabilities to options"?
> 
> What do you all think?

HI @diemol , 

Thanks for the review.

As suggested above, placing code samples will complete the document. To keep the document more precise, I will add some more content.

I will work on this offline..!

Thanks,
Harsha

PR #373, opened 2020-03-24
Driver specific capabilities spanish

jmartinezpoq:
### Description
Translate driver specific capabilities page to spanish

### Types of changes

- [x] Improved translation
- [x] Added new translation (and I also added a notice to each document missing translation)

### Checklist


jmartinezpoq:
Hi @diemol 
Why is the build failing?
Is there anything that we can do to avoid the fail?

Thank you!

diemol:
@jmartinezpoq it is the GitHub API quota that gets hit sometimes, during the process of building the pages some API calls are made, so the public quota is used. We need to figure out a workaround.

PR #446, opened 2020-06-01
Best Handling of Two factor authentication

nainappa:
### Description
This PR includes how to handle 2FA properly in our test environments.

### Motivation and Context
To give more info on worst practices that are being followed in automation.

### Types of changes

- [x] Change to the site (I am attaching a screenshot showing the before and after)
- [ ] Improved translation

### Checklist

After:
--------------------------------------------------------------------------------------------------------------
image


nainappa:
Hi @Harsha509 ,

Thanks for the review. I agree with wording that _"Automating this seamlessly and consistently is a big challenge in Selenium"_ can be modified further. But it is indeed a challenging even with other third party APIs and Twilio like apps as you said. 

Here are my two cents on this.

### Using APIs for TOTP apps:
These TOTP apps generate OTP for every specific duration(eg: 30 sec for google authenticator). We have several APIs which can calcualte the OTP by using our secret key and system time, and we can create a method to fetch this OTP to use in our automation. Below are the issues with this approach.

1.  Consider that you are calling this method at the dying moment of the duration (eg: 29.59sec in google authenticator). This method calculates and gives you the valid OTP. But by the time when you enter this OTP in the application and proceed further, this OTP will not be valid anymore. So, we have consistency issues here.
2.  Companies will not be able to share the secret key with the QE teams as this 2FA would be used across different teams, and in some cases whole company. So, security teams will not accept sharing this.
3.  Using the secret key is always a security concern because, it can be misused or in some cases we might expose in the code unknowingly.

### Twilio like apps for SMS:
Here we will have to first get a number by paying some amount. Then they share a secret key to access our number and fetch the OTP. Below are the challenges in this case.

1.  Though twilio promises realtime delivery of the messages, some times it takes more than few seconds to receive and that OTP might become stale by then.
2.  We are depending on a third party app to get our OTP and security teams may not accept doing this as now we are depending on some other app's security as well.
3.  This is a paid service.
4.  This indeed adds an extra layer on our existing selenium code.

So, please let me know your thoughts.

nainappa:
@diemol Thanks for the review. I made the suggested changes now. Please take a look when you have time.

PR #453, opened 2020-06-05
Black Lives Matter

diemol:
Home page will look like this:
Screen Shot 2020-06-06 at 00 30 29-fullpage


marknoble:
@diemol What is the process for having banners for charities and causes added to the Selenium website front page?

I think the link for IJ.org should be included in the header.  Shall I submit a pull request?  Here's the link that should be added...

https://www.ij.org/support/give-now

diemol:
We do not really have a policy for that, we are still working on setting a better framework for cases like this one.

I believe you can reach us out at selenium@sfconservancy.org or through our [Slack channel](https://www.selenium.dev/support/) to talk about it.

PR #511, opened 2020-09-03
Fix broken edit links

titusfortner:
When you click on the edit links they are broken, looks like they are going to the `trunk` branch, but should be going to the `edit` branch. I'm pretty sure this is the fix, but haven't tested it locally. @diemol can you verify for me? Thanks!


diemol (approved):
Thank you, @titusfortner!

PR #620, opened 2021-01-25
Postmortem report for selenium.dev outage

harsha509:
### Description
Postmortem report for Selenium.dev outage

### Types of changes

- [ ] Improved translation

### Checklist


diemol (changes requested):
Thanks for this, @JS31096!

Perhaps we can sync and add a root analysis section, something like https://en.wikipedia.org/wiki/Five_whys?

PR #675, opened 2021-04-28
Added explanation that 'Relative Locators' is not available in the current stable release.

honjiaxuan:
Added explanation that 'Relative Locators' is not available in the current stable release in docs_source_files/content/webdriver/locating_elements.en.md and all its other language files.

### Description

I added a short paragraph to explain that "Relative Locators" are not available in the current stable release (Selenium 3), anyone that want to use that feature have to install Selenium 4 and latest available version is mentioned as well.

### Motivation and Context

The current documentation didn't tell the users that the stable release they installed is Selenium 3 instead of Selenium 4. But "Relative Locators" only available in Selenium 4.

### Types of changes

- [x] Change to the site (I am attaching a screenshot showing the before and after)
- [ ] Improved translation

### Checklist


luke-hill:
Hi @honjiaxuan 

There appears to be some auto-formatting bug with your IDE that has caused a load of methods to be shifted out of line. Could you check maybe your local editorconfig and maybe try revert those changes?

diemol (changes requested):
This PR creates more maintenance than rather clarifying things. 

I'd be ok with a comment next to the **Selenium 4** text above that says something like "be sure to be on Selenium 4 to use relative locators".

diemol:
@honjiaxuan, would you like to continue with this PR based on the commend I made above?

harsha509:
Closing this as we didn’t hear back from the OP.

PR #681, opened 2021-05-09
Update two_factor_authentication.fr.md

Greavox:
### Description
Translation of some text.

I have a question about it:
What means this: "Disable 2FA if you login from certain IPs. That way we can configure our test machine IPs to avoid this."?

In order for Selelium to test in checking the client IP, it must have those IPs, right? And who is this "we"? Selenium?

### Motivation and Context
I don't use Selenium and I don't know how to use it. But since I understand a little English, I'll take the liberty of translating.

### Types of changes

- [x] Improved translation

### Checklist

### DIFF
--- a/docs_source_files/content/worst_practices/two_factor_authentication.fr.md
+++ b/docs_source_files/content/worst_practices/two_factor_authentication.fr.md
@@ -1,26 +1,25 @@
 ---
-title: "Two Factor Authentication"
+title: "Authentication à double facteur"
 weight: 1
 ---
 
-{{% notice info %}}
-<i class="fas fa-language"></i> Page being translated from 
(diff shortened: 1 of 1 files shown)


diemol:
> "Disable 2FA if you login from certain IPs. That way we can configure our test machine IPs to avoid this."?

Means that the IP(s) from the browser being used to load the application could be allowed to access the application without 2FA. The IPs must be know by the system/application under test. And "we" means the folks who are testing. Selenium is only driving the browser, but for the application under test to work well, the environment should be properly configured.

diemol (approved):
Thank you, @Greavox!
Congrats on your first contribution to the Selenium docs! 🎉

PR #768, opened 2021-09-22
differentiate and improve eplanation for Bidi vs DevTools

titusfortner:
Please review my explanation and make sure that is actually correct. :)

I didn't add code examples or placeholders for all of the APIs we've created, this is just so that we aren't doing what we've been telling other people not to do.

### DIFF
--- a/website_and_docs/content/documentation/support_packages/chrome_devtools.en.md
+++ b/website_and_docs/content/documentation/support_packages/chrome_devtools.en.md
@@ -1,23 +1,28 @@
 ---
-title: "Chrome Devtools"
-linkTitle: "Chrome Devtools"
+title: "Chrome DevTools Protocol"
+linkTitle: "Chrome DevTools Protocol"
 weight: 5
 aliases: ["/documentation/en/support_packages/chrome_devtools/"]
 ---
 
 {{% pageinfo color="warning" %}}
 <p class="lead">
-  While Selenium 4 provides direct access to Chrome DevTools Protocol, it is highly encouraged that
+  While Selenium 4 provides direct access to the Chrome DevTools Protocol (CDP), it is highly encouraged that
 you use the <a href="../../webdriver/bidi_apis">WebDriver Bidi APIs</a> instead.
 </p>
 {{% /pageinfo %}}
 
-Chrome DevTools is a set of web developer tools built directly into the Google
-Chrome browser. The implementation is not designed for testing, nor to have a stable API,
+Many browsers provide "DevTools" -- a set of tools that are integrated with the browser that developers can use to 
+debug web apps and explore the performance of their pages. 
+Google Chrome's DevTools make use of a protocol called the Chrome DevTools Protocol (or "CDP" for short). 
+As the name suggests, this is not designed for testing, nor to have a stable API, 
 so functionality is highly dependent on the version of the browser.
-If there is DevTools functionality that is not yet covered by a Bidi API, though, you can
-use DevTools directly to access properties such as Application Cache, Fetch, Network, Performance, Profiler, 
-Resource Timing, Security and Target CDP domains, etc. Here are a couple examples.
+
+WebDriver Bidi is the next generation of the W3C WebDriver protocol and aims to provide a stable API 
+implemented by all browsers, but it's not yet complete. Until it is, 
+Selenium provides access to the CDP for those browsers that implement it 
+(such as Google Chrome, or Microsoft Edge, and Firefox), 
+allowing you to enhance your tests in interesting ways. Some examples of what you can do with it are given below.
 
 ## Emulate Geo Location
 
--- a/website_and_docs/content/documentation/support_packages/chrome_devtools.en.md
+++ b/website_and_docs/content/documentation/support_packages/chrome_devtools.en.md
@@ -5,16 +5,21 @@ weight: 5
 aliases: ["/documentation/en/support_packages/chrome_devtools/"]
 ---
 
-Selenium 4 alpha versions have much awaited native support for Chrome DevTools 
-Protocol through "DevTools" interface. This helps us getting Chrome Development 
-properties such as Application Cache, Fetch, Network, Performance, Profiler, 
-Resource Timing, Security and Target CDP domains etc.
+{{% pageinfo color="warning" %}}
+<p class="lead">
+  While Selenium 4 provides direct access to Chrome DevTools Protocol, it is highly encouraged that
+you use the <a href="../../webdriver/bidi_apis">WebDriver Bidi APIs</a> instead.
+</p>
+{{% /pageinfo %}}
 
-Chrome DevTools is a set of web developer tools built directly into the Google 
-Chrome browser. DevTools can help you edit pages on-the-fly and diagnose 
-problems quickly, which ultimately helps you build better websites, faster.
+Chrome DevTools is a set of web developer tools built directly into the Google
+Chrome browser. The implementation is not designed for testing, nor to have a stable API,
+so functionality is highly dependent on the version of the browser.
+If there is DevTools functionality that is not yet covered by a Bidi API, though, you can
+use DevTools directly to access properties such as Application Cache, Fetch, Network, Performance, Profiler, 
+Resource Timing, Security and Target CDP domains, etc. Here are a couple examples.
 
-## Emulate Geo Location:
+## Emulate Geo Location
 
 Some applications have different features and functionalities across different 
 locations. Automating such applications is difficult because it is hard to emulate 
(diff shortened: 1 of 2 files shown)


website_and_docs/content/documentation/support_packages/chrome_devtools.en.md
```diff
@@ -5,16 +5,21 @@ weight: 5
...
-Resource Timing, Security and Target CDP domains etc.
-
-Chrome DevTools is a set of web developer tools built directly into the Google 
-Chrome browser. DevTools can help you edit pages on-the-fly and diagnose 
-problems quickly, which ultimately helps you build better websites, faster.
-
-## Emulate Geo Location:
+{{% pageinfo color="warning" %}}
+<p class="lead">
+  While Selenium 4 provides direct access to Chrome DevTools Protocol, it is highly encouraged that
```
shs96c: "the Chrome DevTools Protocol (CDP)"

There are two changes here: first "the" which reads more naturally. Secondly, we're providing the abbreviation next to the definition, so it's okay for you to use "CDP" later in this explanation.

website_and_docs/content/documentation/support_packages/chrome_devtools.en.md
```diff
@@ -5,16 +5,21 @@ weight: 5
...
+you use the <a href="../../webdriver/bidi_apis">WebDriver Bidi APIs</a> instead.
+</p>
+{{% /pageinfo %}}
+
+Chrome DevTools is a set of web developer tools built directly into the Google
+Chrome browser. The implementation is not designed for testing, nor to have a stable API,
+so functionality is highly dependent on the version of the browser.
+If there is DevTools functionality that is not yet covered by a Bidi API, though, you can
+use DevTools directly to access properties such as Application Cache, Fetch, Network, Performance, Profiler,
```
shs96c: The examples are a hostage to the future: as bidi evolves, this will change. Maybe something like:

"If there is DevTools functionality that is not yet covered by a Bidi API, you can use the CDP directly to access that functionality"

website_and_docs/content/documentation/support_packages/chrome_devtools.en.md
```diff
@@ -5,16 +5,21 @@ weight: 5
...
-
-## Emulate Geo Location:
+{{% pageinfo color="warning" %}}
+<p class="lead">
+  While Selenium 4 provides direct access to Chrome DevTools Protocol, it is highly encouraged that
+you use the <a href="../../webdriver/bidi_apis">WebDriver Bidi APIs</a> instead.
+</p>
+{{% /pageinfo %}}
+
+Chrome DevTools is a set of web developer tools built directly into the Google
```
shs96c: I'm wary of confusing people between "devtools" (available on many browsers, including Safari, Firefox, and IE) and "CDP" (a protocol that the chromium-based browsers, and to an extent Firefox, implements) It turns out that Chromium's DevTools are based on CDP, but that's not the case for all browsers.

Perhaps spell that out: "Chrome's DevTools are a set of web developer tools built directly into Chromium-based browsers. They use the CDP to communicate with the browser...."

website_and_docs/content/documentation/webdriver/bidi_apis.en.md
```diff
@@ -5,19 +5,56 @@ weight: 11
...
 ---
 
-In Selenium 4, new Evented APIs were introduced that
-allow users to be able to capture events from the
-browser as they happen rather than using the
-traditional approach of Request/Response that
-WebDriver has used for other APIs.
+Selenium is working with Browser Vendors to create the 
+[BiDirectional WebDriver Protocol](https://w3c.github.io/webdriver-bidi/) 
+as a means to provide a stable, cross-browser implementation of a subset of the functionality found
```
shs96c: The goal isn't to provide a subset of the CDP. The goal is to provide a stable, cross-browser implementation of a bi-directional API designed for browser automation, with testing a specific use case. We want nothing to do with debugging.

website_and_docs/content/documentation/webdriver/bidi_apis.en.md
```diff
@@ -5,19 +5,56 @@ weight: 11
...
+[Chrome DevTools](../../support_packages/chrome_devtools).
+The traditional webdriver model of strict request/response commands will be supplemented with the ability to
+stream events from the user agent to the controlling software via WebSockets, better matching the evented nature of the browser DOM.
 
-Internally WebDriver will create a WebSocket connection
-to the browser for events and commands to be transmitted.
+While this initiative is being developed, Selenium has implemented this API which is currently supported by
```
shs96c: I don't understand this sentence at all.

website_and_docs/content/documentation/webdriver/bidi_apis.en.md
```diff
@@ -5,19 +5,56 @@ weight: 11
...
+Chrome, Edge and Firefox using the Chrome DevTools Protocol. 
+While the underlying implementation may change, the goal is for this API to remain constant.
 
 The following list of APIs will be growing as the Selenium
 project works through supporting real world use cases. If there
 is a missing API, please raise a [feature request](https://github.com/SeleniumHQ/selenium/issues/new?assignees=&labels=&template=feature.md).
 
+## Register Basic Auth:
+
+Some applications require to keep some pages behind an auth and most of the time
```
shs96c: "Some applications make use of browser authentication to secure pages. With Selenium, you can automate...."

website_and_docs/content/documentation/webdriver/bidi_apis.en.md
```diff
@@ -89,30 +126,208 @@ ensure
...
+
+## Listen to console.log events
+
+Listen to the `console.log` events and register callbacks to process the event.
+
+{{< tabpane langEqualsHeader=true >}}
+{{< tab header="Java" >}}
+import org.openqa.selenium.chrome.ChromeDriver;
+import org.openqa.selenium.devtools.DevTools;
+import org.openqa.selenium.devtools.v87.log.Log;
```
shs96c: Use `org.openqa.selenium.devtools.idealized.log.Log` instead. You don't need to tie yourself to a specific browser version.

website_and_docs/content/documentation/webdriver/bidi_apis.en.md
```diff
@@ -89,30 +126,208 @@ ensure
...
+}
+{{< /tab >}}
+{{< tab header="Python" >}}
+# Please raise a PR to add code sample
+{{< /tab >}}
+{{< tab header="CSharp" >}}
+using OpenQA.Selenium.Chrome;
+using OpenQA.Selenium.DevTools;
+using System;
+using DevToolsSessionDomains = OpenQA.Selenium.DevTools.V91.DevToolsSessionDomains;
```
shs96c: I suspect that .Net also makes this safe to use via the idealised API.

website_and_docs/content/documentation/webdriver/bidi_apis.en.md
```diff
@@ -89,30 +126,208 @@ ensure
...
+import org.openqa.selenium.devtools.DevTools;
+import org.openqa.selenium.devtools.v87.log.Log;
+
+public void consoleLogTest() {
+ChromeDriver driver = new ChromeDriver();
+DevTools devTools = driver.getDevTools();
+devTools.createSession();
+
+    devTools.send(Log.enable());
+    devTools.addListener(Log.entryAdded(),
```
shs96c: devTools.getDomains().log().entryAdded()

shs96c (changes requested):

website_and_docs/content/documentation/support_packages/chrome_devtools.en.md
titusfortner: I like it, and I'll reference the browser version limitation again:
> If there is DevTools functionality that is not yet covered by a Bidi API, though, you can
use the CDP directly to access that functionality. Here are a couple examples that work as of
Chrome v93.

website_and_docs/content/documentation/support_packages/chrome_devtools.en.md
titusfortner: Excellent point. I'm also going to rename this page to "Chrome DevTools Protocol." Not sure it is worth updating the file name as well.

website_and_docs/content/documentation/webdriver/bidi_apis.en.md
titusfortner: Yes, this was the part I was getting handwavey at. What about:

> Selenium is working with browser vendors to create the 
[BiDirectional WebDriver Protocol](https://w3c.github.io/webdriver-bidi/) 
as a means to provide a stable, cross-browser API for leveraging the bidirectional
functionality useful for both browser automation generally and testing specifically. 
Before now, users seeking this functionality have had to rely on 
the [Chrome DevTools Protocol](../../support_packages/chrome_devtools), 
with all of its frustrations and limitations.

website_and_docs/content/documentation/webdriver/bidi_apis.en.md
titusfortner: Trying to set expectations for what is working now. how about:

> The BiDi API code in Selenium 4 currently works on Chrome and Edge, with some support for Firefox.

website_and_docs/content/documentation/webdriver/bidi_apis.en.md
titusfortner: I didn't know about this! This is something I don't think is in many of the tutorials I've seen out there.

website_and_docs/content/documentation/webdriver/bidi_apis.en.md
titusfortner: @jimevans is there a better way to reference this? Thanks.

website_and_docs/content/documentation/webdriver/bidi_apis.en.md
titusfortner: Where does the lambda go, then?

website_and_docs/content/documentation/webdriver/bidi_apis.en.md
jimevans: @titusfortner For referencing raw CDP methods? No, there isn’t a better way. But you only need that ugliness if you’re doing something we haven’t wrapped in a friendly bidi api. If what you’re really asking is, “Is there a user-friendly, version agnostic method for monitoring the console logs,” then, yes, there is [a better way](https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/src/webdriver/JavaScriptEngine.cs).

website_and_docs/content/documentation/support_packages/chrome_devtools.en.md
shs96c: How about something like?

> Many browsers provide "DevTools" -- a set of tools that are integrated with the browser that developers can use to debug web apps and explore the performance of their pages. Google Chrome's DevTools make use of a protocol called the Chrome DevTools Protocol (or "CDP" for short). As the name suggests, this is not designed for testing, nor to have a stable API, so functionality is highly dependent on the version of the browser.
>
> WebDriver Bidi is the next generation of the W3C [WebDriver protocol](https://w3c.github.io/webdriver-bidi/) and aims to provide a stable API implemented by all browsers, but it's not yet complete. Until it is, Selenium provides access to the CDP for those browsers that implement it (such as Google Chrome, or Microsoft Edge, and Firefox), allowing you to enhance your tests in interesting ways. Some examples of what you can do with it are given below.
```

File: website_and_docs/content/documentation/webdriver/bidi_apis.en.md (hunk 4 above)
Comment (shs96c) [reply website_and_docs/content/documentation/webdriver/bidi_apis.en.md]: Sounds good, with two minor changes: it's "WebDriver Bidi" not "Bidi WebDriver", and prefer "API that uses" to "for leveraging". I always feel that words like "leverage" are _very_ corporate.

File: website_and_docs/content/documentation/webdriver/bidi_apis.en.md (hunk 5 above)
Comment (shs96c) [reply website_and_docs/content/documentation/webdriver/bidi_apis.en.md]: There is no Bidi implementation yet. We're all CDP at the moment. Firefox is starting to ship the necessary pieces of the nascent spec, though.

How about:

> Because it's a bad idea to tie your tests to a specific version of a specific browser, the Selenium project recommends using WebDriver Bidi wherever possible. However, until the spec is complete there are many useful things that the CDP offers. To help keep your tests independent and portable, Selenium offers some useful helper classes. At the moment, these use the CDP, but when we shall be using WebDriver Bidi as soon as possible!

File: website_and_docs/content/documentation/webdriver/bidi_apis.en.md (hunk 9 above)
Comment (shs96c) [reply website_and_docs/content/documentation/webdriver/bidi_apis.en.md]:
```java
devTools.addListener(
  devTools.getDomains().log().entryAdded(),
  logEntry -> {
    System.out.println("log: %s", logEntry.getText());
  });
```

shs96c (approved):
Ship it!

diemol (approved):

PR #822, opened 2021-11-01
Create Directory for running Java Examples

titusfortner:
@diemol per our discussion on Slack, what do you think of something like this where we put examples in the `example` directory? We can execute these in a github action if we want as well.


diemol (commented):
I think it only makes sense to create this directory and merge this when we are actually using the code in the examples. Not before, as we could leave this half done.

titusfortner:
Yeah, this is the start of what I was going to do for #825 
I mostly just want to know where we should be putting the example code, and if this makes the most sense.

titusfortner:
Moving to draft while reworking this

titusfortner:
Closing in favor of something more comprehensive

PR #823, opened 2021-11-01
Fix errors to allow site to build

titusfortner:
Current files give me this error when trying to build:

Error: add site dependencies: load resources: loading templates: "/Users/titusfortner/code/seleniumhq.github.io/website_and_docs/layouts/partials/footer.html:4:1": parse failed: template: partials/footer.html:4: unclosed action

Not sure if I need to update Hugo or something? Either way, this PR fixes.


diemol (approved):
Maybe it is related to the Hugo version, but we can merge it anyway.

PR #843, opened 2021-11-09
WIP - reorganize the english docs

titusfortner:
This is the full re-org I want to do, but just shows the changes for English, and doesn't update the aliases.

If this is approved, I'll go back and make the changes incrementally with all of the translations and aliases to make it easier to approve/merge.

The proposed content:
* Structures Tree so that each item for a given level matches the others in the list
* Proper Capitalization
* Mostly rearranges existing content, including splitting up existing content where appropriate
* Adds placeholders for new content (IDE, etc)
* The only "new" info in the docs is the IE Driver Server page, which is (mostly) a copy/paste from the wiki (and likely needs updating)


titusfortner:
Subsumed by #861

PR #852, opened 2021-11-15
move webdriver getting started documentation

titusfortner:
The idea is for the top sections to be the different areas of Selenium.

Getting started for Grid should be in the Grid section, Getting started with IDE in IDE section, etc.
This moves the getting started docs to the WebDriver section


titusfortner:
subsumed by #861

PR #861, opened 2021-12-06
reorganize documentation and update titles

titusfortner:
This replaces #843 with a reorganization of everything including all of the translations.

I have plans to consolidate a number of these articles, but for this PR, there aren't any content changes except for the titles.

1. Reorganization puts main components of Selenium in top level
2. Subcategories makes everything same tenses / parts of speech
3. `title` in metadata is a longer and "Sentence case"
4. `linkTitle` in metadata is shorter and "Title Case"
5. `weight` in metadata only uses even numbers for easier adding/removing in future
6. Links were updated as necessary
7. Aliases were updated for each language
8. Hugo link conventions I'm following / trying to follow (because there are lots of ways to get these to work, but also lots of ways they don't work):
    * Linking to a named file in the same branch:  `ref "file.md"`
    * Linking to a named file in lower branch:  `ref "subcategory/file.md"`
    * Linking to an index_file: `ref "subcategory"
    * Linking to a file in higher branch: `ref "/topcategory/file.md"


titusfortner:
Hmm, at the point that translators need to update the titles, perhaps we should go through everything as part of this PR and make sure everything has a description?

diemol (approved):
This is amazing, thank you, @titusfortner!

PR #865, opened 2021-12-07
split out browser code into manageable chunks

titusfortner:
There is a ton of info in the main Browser page, this splits it out into manageable chunks.

Again, not really changing content, just moving it around. A couple of the titles / headers need to be translated, but that's it.


diemol (approved):
Thank you, @titusfortner!

PR #872, opened 2021-12-11
create shortcode for markdown tabs

titusfortner:
I'm tempted to put this in draft because it's not the best way to do this, even though it works.

Current docsy theme forces tabs to have highlighting (https://github.com/google/docsy/issues/550)
Would be nice if we could toggle that off on a per-tab basis.

Since I still can't wrap my head around go/hugo enough to do it the right way, I just copied this from the previous docsy implementation when they didn't have automatic highlighting.

I also added padding and a border, but that's about the extent of my design skills.

### DIFF
--- a/website_and_docs/assets/scss/_tabpane-md.scss
+++ b/website_and_docs/assets/scss/_tabpane-md.scss
@@ -1,0 +1,8 @@
+.tabpane-md {
+  border-style: outset;
+  border-width: thin;
+}
+
+.tabpane-md .tab-pane {
+  padding: 10px;
+}
--- a/website_and_docs/assets/scss/main.scss
+++ b/website_and_docs/assets/scss/main.scss
@@ -8,3 +8,4 @@
 @import "links";
 @import "logo";
 @import "screen";
+@import "tabpane-md";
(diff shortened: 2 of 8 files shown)


titusfortner:
One of the big advantages of using something like this, imo, is that IntelliJ recognizes Ruby comments (`#`) as comments when it's inside 3 ticks, but thinks it is an h1 when it is just inside the tab, so it's much harder to work with the document.

diemol (changes requested):
This looks nice 
image

Is it possible to remove the black border for the tab?

We can merge this, but also, could you please open a PR to docsy? It'd be nice to have this working upstream and then we just use it without having this custom code.

titusfortner:
I'm converting this to a draft, because I really want us to find a way to create a toggle for this behavior and make a PR to docsy. I don't like having to specify all the tabs in the tabpane with this PR, it just feels clunky.

I added the border because without the `highlight` background, it's really hard to see where the tab content ends. We could do an <hr> instead. I set it to be an `outset` border, which makes it more pronounced at the bottom than the top because that's where it is needed.

titusfortner:
Closed in favor of #915

PR #873, opened 2021-12-11
describe element actions

titusfortner:
Moves some of the non-Actions API commands out of Actions API pages and creates a page for describing the 5 basic element actions.

Most of this content is just moved from one place to another.

Also added "Coming Soon" comments for things that need to get added.


diemol (changes requested):
This is a good idea, however I would remove the "coming soon" comments. I would prefer to create issues for that.

titusfortner:
There's more code that needs to be added, and the wheel files are straight placeholders until after Selenium 4.2 is released, but the content for all of this has been overhauled as well.

titusfortner:
I tweaked the alert messages for code & content to allow both generic and specific usage
Replaced all the "coming soon" with alerts requesting help

diemol:
Merged as part of #920

PR #875, opened 2021-12-11
create first script page

titusfortner:
The example code on the main docs page is more complicated than I think we should be showing right off. I've replaced it with a trivial example that's easy to look at, but maybe I've gone too far the other way?

The code to do a Google search I moved to a "first script" page so that we can put the commands in context of what kind of interaction it is. I'm doing this part now and holding off on all other code examples, mostly to provide structure for how I'm organizing other pages in browser & element sections.

### DIFF
--- a/website_and_docs/content/documentation/webdriver/getting_started/first_script.en.md
+++ b/website_and_docs/content/documentation/webdriver/getting_started/first_script.en.md
@@ -1,0 +1,324 @@
+---
+title: "Writing your first Selenium script"
+linkTitle: "First Script"
+weight: 3
+description: >
+    Step-by-step instructions for constructing a Selenium script
+---
+
+Once you have [Selenium installed]({{< ref "install_selenium_library.md" >}}) and
+[Drivers installed]({{< ref "install_drivers.md" >}}), you're ready to write Selenium code.
+
+Everything Selenium does is send the browser commands to do something or send requests for information.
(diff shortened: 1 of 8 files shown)


website_and_docs/content/documentation/webdriver/getting_started/first_script.en.md
```diff
@@ -0,0 +1,203 @@
...
+    // 7. End the session
+    driver.quit()
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+{{% pageinfo color="warning" %}}
+Coming soon:
+* Link to advanced code with a complex form filling example
+  {{% /pageinfo %}}
```
diemol: I would remove this, I see "coming soon" comments as something we might never have :)

diemol (changes requested):
This is nice, thanks!
I left a comment, but aside from that, 👍

titusfortner:
Quick note...

All the code comments have been frustrating me for a few reasons:
1. They are duplicated everywhere in our examples
2. They don't lend themselves well to translation
3. My  IDE sees Ruby & Python comments `#` as headers of `*.md` files, rather than code comments, so I can't collapse sections in a way that makes it easy to work with.

My solution here was to have a bunch of different tabbed sections, explaining each item separately. Ideally we can refer back to this example or some such rather than duplicating things everywhere. The full script (sans comments) is at the end.

PR #877, opened 2021-12-12
Move wiki content to docs

titusfortner:
Looking to clean up the wiki a bit by moving content to documentation.

There were 65 articles. I just outright deleted 3 that didn't add any value.
I think we can just delete this one rather than move it: https://github.com/SeleniumHQ/selenium/wiki/Continuous-Integration

This PR contains the content of 11 others. Many of our wiki articles are 10 years old, so most of the content is going into the "Legacy" documentation.

I added "Previously located" link for where I'm copying the article from. That probably isn't necessary.
The thought was to update the wiki with a link to where the content has move, like this:
https://github.com/SeleniumHQ/selenium/wiki/InternetExplorerDriver

So, I managed to address about a quarter of the wiki articles, which isn't nothing, but was definitely tiring.

Oh, and I don't think we need to translate anything that's in Legacy section. Maybe should for the Test Practices ones?


diemol:
No, I do not think Legacy needs translations.

diemol (approved):
Thank you, @titusfortner!

Will you remove the Wiki page you mentioned and also add links in the Wiki to the new pages in the docs?

PR #902, opened 2021-12-22
Alerts

titusfortner:
@ddavison did I lose any of the stuff you added in dismissible banner with this implementation?

This is to consolidate the banner / alert code to make it easy to add custom short codes with minimal end user annotation. Examples for the new `alert-code` and `alert-content` are in the `contributing.en.md` file

@diemol this is my suggested alternative to the "Coming Soon" comments I made in the other PRs. Let me know if I should tweak this language. 

If it looks ok, I'll update my other PRs with this.


ddavison (approved):
> @ddavison did I lose any of the stuff you added in dismissible banner with this implementation?

i didn't check out your code specifically, but it looks great!  It doesn't look like we lost anything content-wise.  Nice work!

PR #904, opened 2021-12-23
store meeting agenda and transcripts in separate place on site

titusfortner:
I don't think either the wiki or the blog is the ideal location for archiving meeting details.

I created a new Hugo "section" for "meetings." I barely know what I'm doing, but what I ended up with looks ok to me.

New card on https://www.selenium.dev/project/
which links to the new list page: https://www.selenium.dev/meetings/

I couldn't figure out how to make this work nested (e.g., https://www.selenium.dev/project/meetings/), but I think it's ok without that.

baseof.html is a copy of the blog one, except without the categories
list.html is a straight copy from blog
single.html I copied from hugo docs because whatever the default was, it wasn't displaying the title on the pages

Also I copied all of the relevant ones from blog to here, I'm not sure it makes sense to delete them from blog, but I do want them all in one spot.

Please suggest fixes for any names/descriptions in the frontmatter, my brain is limited right now.

### DIFF
--- a/website_and_docs/content/blog/2020/public-project-meeting-20200409.md
+++ b/website_and_docs/content/blog/2020/public-project-meeting-20200409.md
@@ -9,6 +9,9 @@ description: >
   As an initiative to bring more transparency to the Selenium project...
 ---
 
+{{< dismissible-banner title="Archived meeting minutes" alert="note" color="blue" >}}
+All meeting minutes can now be found [here](/meetings).
+{{< /dismissible-banner >}}
 
 As an initiative to bring more transparency to the Selenium project, we are now holding bi-weekly public project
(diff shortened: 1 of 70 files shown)


diemol (changes requested):
I think this makes sense and the PR looks good. However, I do not see the meeting minutes being deleted from the blog section. Could you please remove them? Otherwise this PR adds zero value :trollface:

titusfortner:
The primary goal was to move things out of the wiki — https://github.com/SeleniumHQ/selenium/wiki/TLC-Meetings
I figured at the point we were putting them somewhere specific, they should all be there.
I didn't want to delete old blog posts because deleting old blog posts is generally frowned upon, and the duplication seems minor.

titusfortner:
Talked to my journalist better half and she says we should put a link to the new location at the top of the blog posts so it doesn't look like we're deleting previously published content, so I'll do that.

diemol:
> Talked to my journalist better half and she says we should put a link to the new location at the top of the blog posts so it doesn't look like we're deleting previously published content, so I'll do that.

You can also add redirects. Much easier.

titusfortner:
bunch of updates to make this work properly

diemol (approved):
Thank you, @titusfortner!

PR #908, opened 2021-12-28
doc: Correct instructions to alter .profile $PATH

SteveParson:
### Description
Currently, instructions are provided with the intent to add the driver path to the PATH environment variable. As written, the instructions do nothing. The commands were modified make the intended changes.

### Motivation and Context
Users that aren't familiar with the PATH environment variable may get confused when the changes weren't persisted as intended.

### Types of changes

- [x] Change to the site (I have NOT double-checked the Netlify deployment, and my changes look good)
- [ ] Improved translation

### Checklist

### DIFF
--- a/website_and_docs/content/documentation/webdriver/getting_started/install_drivers.en.md
+++ b/website_and_docs/content/documentation/webdriver/getting_started/install_drivers.en.md
@@ -141,7 +141,8 @@ echo %PATH%
 
 {{< tabpane  >}}
 {{< tab header="Mac / Linux" >}}
-export PATH=$PATH:/opt/WebDriver/bin >> ~/.profile
+echo 'PATH="$PATH:/opt/WebDriver/bin"' >> ~/.profile
+source ~/.profile
 {{< /tab >}}
 {{< tab header="Windows" >}}
 setx PATH "%PATH%;C:\WebDriver\bin"
--- a/website_and_docs/content/documentation/webdriver/getting_started/install_drivers.ja.md
+++ b/website_and_docs/content/documentation/webdriver/getting_started/install_drivers.ja.md
@@ -139,7 +139,8 @@ echo %PATH%
 
 {{< tabpane  >}}
 {{< tab header="Mac / Linux" >}}
-export PATH=$PATH:/opt/WebDriver/bin >> ~/.profile
+echo 'PATH="$PATH:/opt/WebDriver/bin"' >> ~/.profile
+source ~/.profile
 {{< /tab >}}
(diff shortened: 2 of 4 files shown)


titusfortner (changes requested):
Ah, yes, current instructions are correct for bash not zsh, which is the "new" default for mac. Probably need to make that distinction clear somehow here.

SteveParson:
> Ah, yes, current instructions are correct for bash not zsh, which is the "new" default for mac. Probably need to make that distinction clear somehow here.

Thanks for your reply! I tested my changes in a bash shell, not zsh. I have not used zsh, but I don't think the existing instructions would be correct for that either.

titusfortner:
Oh, I read it backwards. Echo is for bash, export is for zsh.

SteveParson:
The problem with the existing code is that it adds to the local path, but it doesn't add anything to `.profile` as intended.

For example:
```shell
export PATH=$PATH:/opt/WebDriver/bin >> ~/.profile
```

The first part of this command redefines the PATH variable for this specific session. In other words, if the path is currently `/bin:/usr/bin`, this command will replace it with `/bin:/usr/bin:/opt/WebDriver/bin`. **This will not persist into the next shell session**

The next part of this command is intended to deal with persistence of the path. 
Note the redirect `>>`. This causes the output of the previous command to be appended to `~/.profile` as specified. 
Usually the intent of appending to `~/.profile` is to add a line to it. In that case, using `echo` (as in my PR) will print the line to stdout, which will be appended to `~/.profile` by means of  ` >> ~/.profile`
Unfortunately, there is no output from the `export` command, and thus nothing will be appended to `~/.profile`.

Now, to make things more complicated, there are multiple places to add to the user's PATH variable. For instance, on my local install of Ubuntu 18.04, adding to `~/.profile` is the best place to add it by default. However, if the user has `~/.bash_profile` or `~/.bash_login` defined, then those places will be used instead. `~/.bashrc` is also another place that these kind of things can be added.

It is my understanding that in MacOS running zsh, then `~/.zshenv` or `~/.zshrc` are the default places to add to the path, however I have not personally verified this.

titusfortner:
This is what I'm proposing to address this... Separate Bash & Zsh tabs, and I verified these work.

(This code is buried inside #920)

https://github.com/SeleniumHQ/seleniumhq.github.io/blob/e84b02fdec51ac286c4d7666a1902cf2d34b9918/website_and_docs/content/documentation/webdriver/getting_started/install_drivers.en.md?plain=1#L171-L219

SteveParson:
Even better!

PR #915, opened 2022-01-07
implement a toggle for docsy tabpanes so that it doesn't automatically add code

titusfortner:
This replaces #872 

I figured out how to toggle instead of replace, this is a much easier way to do this. 
@diemol I'm waiting to PR it to Docsy based on your feedback.

You can see examples of everything in the new style guide page I made.

### DIFF
--- a/website_and_docs/assets/scss/_tabpane.scss
+++ b/website_and_docs/assets/scss/_tabpane.scss
@@ -1,5 +1,11 @@
-div.nocode {
+.tab-pane .nocode {
   border-bottom: 1px solid #dee2e6;
   margin: 2rem 0 2rem 0;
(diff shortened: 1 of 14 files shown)


titusfortner:
Added a new commit to put a GitHub link at the bottom of a tab in auto-code-block mode

diemol (changes requested):
I think we can merge this, but there seems to be a conflic I would prefer you to fix it so we can merge. Thanks, @titusfortner!

titusfortner:
Ok, I rebased this, adding the other new content to the new Styles documentation page.
Also slightly better CSS than what I had before

diemol (approved):

PR #920, opened 2022-01-09
Example code

titusfortner:
This PR is several things:

1. A new directory (`/examples`) to contain repos in Java, Ruby, .NET and Python for putting code examples
2. A real-world example of what the shortcode in #915  can do (this PR relies on that PR)
3. A re-working of the Getting Started pages with easier to process code examples

Note the ability to mix between code and text in the tabs
Note the "See full example on GitHub" links (they are working, but we need to decide how we will keep them synchronized)
I added an implicit wait to first script because of stupid Google timing issues.

* I don't know how to create a basic JS project or Kotlin project, so there is no repo and no links to those code snippets, and while I think they work, I'm not an expert
* We can add a github action to run this code if we like this setup

This PR is going to need some rebasing still, but I need some feedback on it.


titusfortner:
Ah, I started to make links to the browser specific functionality in open browser and only did it for Chrome. I'll fix that.

titusfortner:
This now includes #873 because the steps in the getting started guide don't make sense without those new links.
I incorporated all the newest updates from other users, and rebased against the translations.

diemol (approved):
Thank you, @titusfortner!

PR #996, opened 2022-04-20
fixed hardcoded driver location snippet for Python

Seseikelele:


diemol (commented):
Thanks for this PR, @Seseikelele, however this usage has been deprecated

https://github.com/SeleniumHQ/selenium/blob/932ac9901a790e0182d7f1af1c44ad3599ffd720/py/selenium/webdriver/chrome/webdriver.py#L56

Also, for future reference, it'd  be nice to have a more complete PR description to understand the motivation behind it.

PR #1010, opened 2022-05-16
Wheel examples

titusfortner:
This is Draft because:
* the gh-codeblock is temporarily pointing to the wrong branch for demo purposes & needs to be changed back before merging
* Depends on merging https://github.com/SeleniumHQ/selenium/pull/10650

This PR is to show off demo code for what scrolling implementation looks like with the above PR

Also, probably should figure out if we like the badges and do them the right way with shortcodes and such if we like them. @diemol what do you think of the "badges" in this PR?

### DIFF
--- a/examples/java/src/test/java/dev/selenium/actions_api/WheelTest.java
+++ b/examples/java/src/test/java/dev/selenium/actions_api/WheelTest.java
@@ -1,0 +1,103 @@
+package dev.selenium.actions_api;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.openqa.selenium.By;
+import org.openqa.selenium.JavascriptExecutor;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.chrome.ChromeDriver;
+import org.openqa.selenium.interactions.Actions;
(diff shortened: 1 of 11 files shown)


titusfortner:
Oh, also, the JavaScript & the Kotlin aren't implemented for this because I wasn't sure what those needed to look like.

titusfortner:
Do I need to move this out of draft to get it to deploy a preview?

examples/java/src/test/java/dev/selenium/actions_api/WheelTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private boolean inViewport(WebElement element) {
+
+        String script =
+                "for(var e=arguments[0],f=e.offsetTop,t=e.offsetLeft,o=e.offsetWidth,n=e.offsetHeight;\n"
+                        + "e.offsetParent;)f+=(e=e.offsetParent).offsetTop,t+=e.offsetLeft;\n"
+                        + "return f<window.pageYOffset+window.innerHeight&&t<window.pageXOffset+window.innerWidth&&f+n>\n"
+                        + "window.pageYOffset&&t+o>window.pageXOffset";
+
+        return (boolean) ((JavascriptExecutor) driver).executeScript(script, element);
```
alaahong, line 101: executeScript will return Boolean as Object, seems no need to convert as boolean?

diemol:
I like the badges idea, but there is no need to implement this on our own. Docsy uses Bootstrap, so you can use their Badges (https://getbootstrap.com/docs/4.6/components/badge/) and combine it with a Tooltip (https://getbootstrap.com/docs/4.6/components/tooltips/). 
That could be embedded in a shortcode, definitely.

titusfortner:
Getting rolled into #1031

PR #1092, opened 2022-07-28
Added more details on moveByOffset to make it clear which directions the mouse moves

### LINKED ISSUE #1087: [🐛 Bug]: Actions API moveByOffset is not clear on the directions of movement with the X,Y input arguments
### What happened?

I had a lot of confusion when I was first learning to use moveByOffset. To fully understand, I did a series of small experiments to understand how the movement around the screen occurs.  I now understand that a positive X value moves the mouse to the right of its current position, while a positive Y value moves down.  

To make it more clear, I propose adding the following note:

```
Note that the first argument X specifies to move right when positive, while the second argument Y specifies
to move down when positive. So moveByOffset(30, -10) moves right 30 and up 10 from the current mouse position.
```

as well as:

```
Rules:
- positive X moves right
- positive Y moves down.
- negative X moves left.
- negative Y moves up.
```

### What browsers and operating systems are you seeing the problem on?

I was using Chrome latest version, and Mac OS M1.

jamesmortensen:
Fixes #1087

I felt the documentation for moveByOffset wasn't clear in terms of how the mouse moves based on X and Y inputs.  So I propose this change to make it clear how a positive X and a positive Y moves the mouse pointer.

### Description
This is a simple addition to the moveByOffset documentation.

### Motivation and Context
I myself experienced some confusion on this. I incorrectly assumed a positive Y would move the mouse up.  But after experimenting I realized a positive Y moves the mouse down.  So I documented that in this PR.  I have also seen other QA engineers struggle with the mouse portion of the Actions API. I hope these changes will make it more clear.

### Types of changes

- [ ] Improved translation

### Checklist

- [ X] I have read the [**contributing**](https://selenium.dev/documentation/en/contributing/) document.
- [ X] I have used [hugo](https://gohugo.io/) to render the site/docs locally and I am sure it works.

### DIFF
--- a/website_and_docs/content/documentation/webdriver/actions_api/mouse.en.md
+++ b/website_and_docs/content/documentation/webdriver/actions_api/mouse.en.md
@@ -310,6 +310,10 @@ If the mouse has not previously been moved, the position will be in the upper le
 corner of the viewport.
 Note that the pointer position does not change when the page is scrolled.
 
+Note that the first argument X specifies to move right when positive, while the second argument
+Y specifies to move down when positive. So `moveByOffset(30, -10)` moves right 30 and up 10 from
+the current mouse position.
+
 {{< tabpane disableCodeBlock=true height="3" >}}
     {{< tab header="Java" >}}
         {{< gh-codeblock path="examples/java/src/test/java/dev/selenium/actions_api/MouseTest.java#L153-L155" >}}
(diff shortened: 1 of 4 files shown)


diemol (changes requested):
Can you please add this to the other `mouse.*.md` files?

diemol (approved):
Thank you, @jamesmortensen!

PR #1121, opened 2022-08-15
Unify first script example (code and doc) between different languages

bonigarcia:
### Description
This PR unifies the logic of the "first script" page between the different binding languages.

### Motivation and Context
At this moment, there are slight differences between the "first script" example. For instance, some uses duckduckgo as SUT, and others not. Duckduckgo is not very convenient of testing in CI since it eventually has a slow response.

### Types of changes

- [X] Change to the site (I have double-checked the Netlify deployment, and my changes look good)
- [X] Code example added (and I also added the example to all translated languages)
- [ ] Improved translation

### Checklist

- [X] I have read the [**contributing**](https://selenium.dev/documentation/en/contributing/) document.
- [X] I have used [hugo](https://gohugo.io/) to render the site/docs locally and I am sure it works.


diemol (approved):
Thank you, @bonigarcia!

PR #1128, opened 2022-08-18
Modification to PLC membership governance language

wcmcgee:
**A PR well described will help maintainers to quickly review and merge it**

Avoid large PRs, help reviewers by making them as simple and short as possible.

### Description
Modification to PLC membership language.

### Types of changes

- [ ] Improved translation

### Checklist

### DIFF
--- a/website_and_docs/content/project/governance/_index.html
+++ b/website_and_docs/content/project/governance/_index.html
@@ -396,7 +396,7 @@ aliases: ["/governance/"]
 
       <p class="pb-2">
         PLC seats are not time-limited. Only one person on the PLC may be affiliated with
-        any given employer at a time. The PLC member count should be always
+        any given employer at a time, with one exception; the Project Lead may be added to the PLC and is exempt from this rule. The PLC member count should be always
         odd so that no votes are ever tied, an ideal minimum number is 5. PLC members
         should also be active members of the community.
       </p>


AutomatedTester:
I am against this change for the following reasons: 

* The selenium project does not need a Project Lead. The project has for years, even when SImon was BDFL, worked mostly from consensus, especially from the TLC.
* There is no documentation on how a Project Lead is created and what that process might be and how long that role exists for.
* The Single person from a company is in place to prevent one company from exerting too much influence on the project. It was originally designed to prevent Google from taking over the project. I would like to maintain that mechanism as I could, as a bad actor, push more BrowserStack engineers into things if I wanted to change the direction of the project.

titusfortner:
This PR does not specify who the Project Lead would be, or that one has to exist, just that if there were one, they wouldn't be subject to the one-person-per-company rule to participate on PLC. The goal was to explicitly avoid blanket changing that rule and only making one exception. I do not want there to be a Project Lead that wasn't agreed on by consensus of the whole community, and there's nothing in this PR that indicates it would be so. Adding clarity around how one, if they were to exist, would be chosen and what their responsibilities would be and if there is something like "term limits" makes sense to me.

Part of the issue here is that Selenium leadership got stuck after Simon left and Jim stepped down because the combination of the remaining PLC members and the SFC were not working well together (for *reasons). The one person on this project who was effectively driving things forward was not allowed to adequately push for needed changes because of the employer rule. Because of our new PLC members, this is less of an issue, and I believe that some of the restrictions placed on non-PLC members participating in meetings, etc are being lifted. As such, I don't think having someone in this position to drive changes is needed at this point. At the same time, we de facto have a project lead, and it would make sense for multiple reasons to officially recognize that.

Selenium has always been a high trust project. We don't remove commit bits, we generally believe that people will do the right things if they haven't proven otherwise. That's the reason I've enjoyed participating in and contributing to it. To be honest, I'm much less interested in participating in the project if we move to some form of lower-trust governance. A Bad Actor taking control of the project would require multiple levels of failure that, to my mind, are extremely unrealistic. If we maintain 5+ PLC members the way the governance recommends, that means at best there will be 2 people from any company, both of which were elected by community consensus, and one of which is doing so much work that we are willing to consider them the current face of the project. I don't think people involved in this project are very shy about sharing their opinions. If there is an issue with how things are being directed or that there is a decision that can be perceived as unfair, the chance that it goes by without comment is effectively zero.

Finally, I would *love if vendors focused on the project more. Yes, throw multiple developers, scratch the itches that are holding back commercial acceptance of our project; provide a Product person to help coordinate issues and backlogs and run meetings; add a tech content person to get our docs better structured and written, or a community person to oversea an Ambassador program, or create a Support Sheriff initiative. Selenium can use so much more help than it's getting I think the objection here is concerned about the exact opposite of our current problem.

diemol:
Closing this PR for the time being as this still needs more internal reviewing.

PR #1142, opened 2022-08-23
Fix-correction-in-"FInding-web-elements"

jithuraj:
Fix-correction-in-"FInding-web-elements"

In [Evaluating a subset of the DOM](https://www.selenium.dev/documentation/webdriver/elements/finders/#evaluating-a-subset-of-the-dom) section, classes are accessed as ids.


diemol (approved):
Good catch, thank you, @jithuraj!

PR #1152, opened 2022-08-28
highlight what the documentation applies to

titusfortner:
At the Test Automation Summit it was suggested that it was not clear what version the example code applies to.

Is this marking sufficient? It's not ideal because I'm overriding the default docsy breadcrumbs template, and if anything other than "documentation" wants to use breadcrumbs, it'll cause an issue.

I did figure out how to make a note that the "Legacy" section is not going to have valid examples.


diemol (approved):
This looks good enough to me.

Thank you, @titusfortner!

PR #1154, opened 2022-08-31
Fix tabs

titusfortner:
This removes our (my) custom implementation of tabpanes, restores us to the docsy official ones.

Take a look at the changes to the Style page to understand the new requirements.

I'm going to move this into draft because I'm not sure I've found all the issues with `%` vs `<` syntax.


titusfortner:
merged directly

PR #1157, opened 2022-09-05
automatically set the height of the tabs across the languages

titusfortner:
I was doing this manually before with "height" values. This is my attempt to do it automatically. @harsha1502 let me know if there's better JS for this. I'm not sure why docsy was doing `$('#' + element.id).tab('show')` and why I could only get it to work with `click()`.

If this looks good, I'll PR it to Docsy as well and see what they think.


titusfortner:
hmm, this doesn't work if a language isn't already set, so needs more tweaking.

titusfortner:
Ok, fixed the issue. Feedback appreciated.

titusfortner:
This had to get reverted because it doesn't consistently work for some reason.

PR #1166, opened 2022-09-14
Even more reorganizations

titusfortner:
Oof another big reorganization.

I need a place to put Firefox Add-on examples, which means a browser specific functionality section.
If we have that, then it makes sense to move browser specific info/execution from `getting_started/open_browser` (which is already not doing a lot to help new users in getting started section) and browser specific capabilities to the same place, call it "browsers"
But what to do about shared capabilities? Move a bunch of driver specific things to the driver section, including placeholders to discuss Service classes, Listeners & Command executors / http library functionality
Now it doesn't make sense to have a "Browsers" section and a "Browser" section, and what does "Browser" even mean? (it was a bad choice of names by me last time), so let's rename that to "Interactions"
Might as well move Virtual Authenticator into Interactions directory & RemoteWebDriver into drivers directory to tidy things.
Quick change of order on the left to highlight content that is viewed more frequently in search result data...

whew.


pujagani (approved):
LGTM
Thank you! @titusfortner

harsha509 (approved):

PR #1172, opened 2022-09-19
4 5 new

titusfortner:
This is examples and documentation for the new 4.5 features for installing unsigned firefox addons and updated select list logic. As such, some of these tests should fail until we release and update Selenium 4.5


titusfortner:
This was already merged

PR #1215, opened 2022-11-05
Added Python get_attribute example

Denperidge:
A simple contribution! Saw that no example was provided for Python, so I made one based on the Java one.


diemol (approved):
Thank you, @Denperidge!

PR #1217, opened 2022-11-07
Remove driver managers

titusfortner:
This removes all extra references to driver managers and tidies up examples and references to examples.

It also gets all example code passing for Se 4.6.

Further optimization possible once Internet Explorer is included in Selenium Managers


diemol (approved):
Thank you, @titusfortner!

PR #1248, opened 2022-11-27
Docs: Update "Cookie" article

DKozachenko:
### Description
Added  working JS code example in "Cookie" article for all languages.

### Motivation and Context
In the current documentation, the "JavaScript" tab shows some obscure html code.

### Types of changes

- [ ] Improved translation

### Checklist

### DIFF
--- a/website_and_docs/content/documentation/webdriver/interactions/cookies.en.md
+++ b/website_and_docs/content/documentation/webdriver/interactions/cookies.en.md
@@ -89,7 +89,18 @@ ensure
 end
   {{< /tab >}}
   {{< tab header="JavaScript" code=false >}}
-{{< gh-codeblock path="/examples/javascript/test/browser/cookies.spec.js#L13-L18">}}
+import { Builder, Browser } from 'selenium-webdriver';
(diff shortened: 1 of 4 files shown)


titusfortner:
Thanks for bringing this to our attention!

We want to move all of our code examples into github, and we're using hugo to dynamically generate that code for display on the pages. So the code already exists here: https://github.com/SeleniumHQ/seleniumhq.github.io/blob/trunk/examples/javascript/test/browser/cookies.spec.js

I just referenced it wrong the last time I updated things. :( 
When any tab uses the gh-codeblock, the `tabpane` needs to include `code=false`, and any tab that still has code needs to use `code=true`. The cookies page here is trying to use `code=false` on only the tab, which is wrong.

I'll work on the fix now. Thanks again for your code proposal, though, and please continue to help us with issues if you find them. We also have a #selenium-docs channel in Slack if you want to discuss anything - https://www.selenium.dev/support/#ChatRoom

titusfortner:
Fixed with - 0f8ef8b22c1d10a458a3f8ce51c1bf3ebdf15399

PR #1254, opened 2022-12-02
Updates for Selenium 4.7

titusfortner:
Running tests to make sure everything is actually correct.
If the tests pass, still need to update the documentation links


titusfortner:
merged

PR #1271, opened 2022-12-24
Add Troubleshooting section

titusfortner:
### Description
* Add troubleshooting section
* Move current exceptions page to troubleshooting/errors page
* Alias current exceptions and older exceptions pages to errors page
* Rewrite errors page
* Add Logger page with Ruby information

### Motivation and Context

I just wanted to put information up about Ruby logging.
I didn't see a place for it, and I thought a `Troubleshooting` section would make sense. Then I remembered that
we have that weird exceptions page that is linked to in the Java code. In the course of moving that over, I realized
that the content is... really bad. It was created by someone in 2015 and no one has evaluated or updated it since.
So that was revamped.


p0deje (commented):
I think it's a great start, but what about localization? I see plenty of files that likely should be translated into other languages, while they are in English.

titusfortner:
We have designated translators for each language who create updates when they have time. The current process is to start with everything in English.

harsha509 (approved):

diemol (approved):
This is great, @titusfortner! When I moved the content to the new site I had this in mind, and it makes me very happy you took the time to organize all this.

PR #1302, opened 2023-02-01
Test page scrolling

galgal12345:
This code opens a Firefox browser, navigates to a desired page, and scrolls down using the arrow keys.   
    The code then verifies that the page has been scrolled by checking that the page height has increased.
    

### Types of changes

- [ ] Improved translation

### Checklist

- [ ] I have read the [**contributing**](https://selenium.dev/documentation/en/contributing/) document.


diemol (commented):
Where would you link this example in the docs? Any reason why you did not use the same page used for the other tests?

titusfortner:
This is an example for how to send a key with actions class. That belongs in the keyboard examples file. 

I don't think we want to create examples for the behavior of every specific key? But we could. 

As far as scrolling goes, we want to encourage it to be always done with the scroll wheel going forward.

diemol:
Closing this based on the last comment.

PR #1305, opened 2023-02-09
Fix em dashes in "--ext" examples

tesueret-msft:
One can guess how I found out that the "--ext" argument examples turned out to be invalid command lines.  :)

### Description

The "--ext" example command-lines in this page previously included an em dash, instead of a hyphen, which caused the commands to be unparseable/invalid.

### Motivation and Context

In my own upgrade to a recent Selenium Server, I copied-and-pasted an "--ext" example, inadvertently breaking my startup commands (which began printing help text instead of executing selenium).

### Types of changes
- [ ] Improved translation

### Checklist

Edits were done within GitHub's live text editor.

### DIFF
--- a/website_and_docs/content/documentation/grid/getting_started.en.md
+++ b/website_and_docs/content/documentation/grid/getting_started.en.md
@@ -239,17 +239,17 @@ The jar file can be downloaded directly from [repo1.maven.org](https://repo1.mav
 and then start the Grid in the following way:
 
 ```bash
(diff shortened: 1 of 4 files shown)


titusfortner:
Good catch! Can you fix it in the other 3 language files as well? Thanks.

tesueret-msft:
> Good catch! Can you fix it in the other 3 language files as well? Thanks.

Oh good point, absolutely :)

PR #1315, opened 2023-02-23
GitHub action workflow scripts: bump actions to their latest versions

deining:
This PR bumps actions used inside GitHub action workflow scripts to their latest versions, thus avoiding deprecation warnings when running the scripts.


diemol (approved):
Thank you, @deining!

PR #1317, opened 2023-02-25
Add missing link text: Chat Room

peterhauke:
Add missing link text "Chat Room" which is the same as the link text at the end of the next section.


harsha509 (approved):
Thank you @peterhauke !

PR #1320, opened 2023-02-28
Adding github action for link checking

deining:
This PR adds a html lint action to the GitHub workflow.
I leave fixing of reported errors as an exercise to you 😁.

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -1,0 +1,54 @@
+# .github/workflows/link-check.yml
+# Run hyperlink link checker on generated HTML output
+name: Link check with hyperlink
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+    - trunk
+  pull_request:
+
+jobs:
+  htmltest:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+
+      - name: Set up Hugo
+        uses: peaceiris/actions-hugo@v2
+        with:
+          hugo-version: "0.110.0"
+          extended: true
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18.14.2'
+          cache: 'npm'
+          # The action defaults to search for the dependency file (package-lock.json,
+          # npm-shrinkwrap.json or yarn.lock) in the repository root, and uses its
+          # hash as a part of the cache key.
+          # https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data
+          cache-dependency-path: '**/package-lock.json'
+
+      - run: npm ci
+        working-directory: website_and_docs
+      - run: hugo --destination $GITHUB_WORKSPACE/website_and_docs/public
+        working-directory: website_and_docs
+
+      - name: Link check
+        continue-on-error: true # <- If set to false, run fails with broken links
+        uses: untitaker/hyperlink@0.1.27
+        with:
+          args: website_and_docs/public/ --check-anchors
+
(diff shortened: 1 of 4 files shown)


.github/workflows/htmltest.yml
```diff
@@ -0,0 +1,59 @@
...
+jobs:
+  htmltest:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # Fetch all history for .GitInfo and .Lastmod
```
diemol: I don't think we need to check the whole history to check links.

diemol (changes requested):
Thank you, @deining! Left a minor comment.

.github/workflows/htmltest.yml
deining: > I don't think we need to check the whole history to check links.

That's very true. I removed the with clause, this certainly will make checkout faster.
You may have to adapt the link checker's configuration file according to your needs, too.

.github/workflows/htmltest.yml
```diff
@@ -0,0 +1,57 @@
...
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+
+      - name: Set up Hugo
+        uses: peaceiris/actions-hugo@v2
+        with:
+          hugo-version: "latest"
```
diemol: I missed this. I prefer to pin the Hugo version and upgrade when we have time to check on things. What is your experience with this?

deining: > I missed this. I prefer to pin the Hugo version and upgrade when we have time to check on things. What is your experience with this?

Yes, that is of course the better setting in your scenario. Changed now.

diemol (approved):
Thank you, @deining!

diemol:
It does seem the job is failing and I cannot see why. Do you have any successful execution to see how the results look like? @deining

diemol:
It also seems to take 1.5 hours to run? That is a bit too much when pushing commits and opening PRs

deining:
> It does seem the job is failing and I cannot see why. Do you have any successful execution to see how the results look like? @deining

Have a look here:

https://github.com/deining/docsy/actions/runs/4281770427

deining:
> It also seems to take 1.5 hours to run? That is a bit too much when pushing commits and opening PRs

Definitely!

I see several possibilities to bring run time down:

```
We store temporary files in tmp/.htmltest by default.
This contains a log of output and a cache of external links,
speeding up subsequent runs.
```

There is also a `-s` switch:

```
  -s, --skip-external          Skip external link checks, may shorten execution
                               time considerably.
```

I would start by running [`htmltest`](https://github.com/wjdp/htmltest/releases) locally and fix the reported issues. Once alle issues ae fixed, I would run the `htmltest` workflow again and is run time is acceptable for you.

I also experimented with [hyperlink](https://github.com/untitaker/hyperlink) which is blazingly fast, but perform internal link checks only. If you want to give it a try, have a look at this [wokflow file](https://github.com/deining/docsy-example/blob/deploy-to-github-pages/.github/workflows/link-check.yml) I authored.

diemol:
The thing is that I am really short of time, we are only a few working on this project and adding this would increase the workload because we need to have this working well so it does not block PRs. Seems I need to leave this open and revisit whenever we have time. Sorry. 🥲

deining:
I switched the engine, the workflow now uses [hyperlink](https://github.com/untitaker/hyperlink) as link checker. Link check lasts 5 sec only, only internal links are checked.

diemol (approved):
Thank you, @deining!

I think, at least for now, we care more about our internal links since we do not use extensively external links.

PR #1321, opened 2023-02-28
[js] Adding logger documentation

TamsilAmani:
### Description

Documentation on how to use Logging feature in JS binding. Formatting documentation for Ruby and JS.

### Motivation and Context

https://github.com/SeleniumHQ/selenium/issues/10944

### Types of changes

- [ ] Improved translation

### Checklist

### DIFF
--- a/website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
+++ b/website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
@@ -78,23 +78,23 @@ JS has implemented logging APIs. It has also added LogManager to maintain a coll
 
 ### 1. Getting a logger can be done in two ways:
 
-```
+```js
 const logging = require('selenium-webdriver/lib/logging')
 ```
 
 First method using manager
-```
+```js
 mgr = new logging.LogManager()
 logger = mgr.getLogger('')
 ```
 
 Second method without manager
-```
+```js
 logger = logging.getLogger('')
 ```
 
 Loggers class uses hierarchical, dot-delimited namespaces (e.g. "" > "webdriver" > "webdriver.logging"). For example, the following code will give a hierarchy of loggers:
-```
+```js
 logger = mgr.getLogger('foo.bar.baz')
 ```
 
@@ -105,12 +105,12 @@ logger = mgr.getLogger('foo.bar.baz')
 
 The basic levels for JS loggers are: `OFF`, `SEVERE`, `WARNING`, `INFO`, `DEBUG`, `FINE`, `FINER`, `FINEST`, `ALL`.
 To log messages at the given level pass the level in the `log()` method:
-```
+```js
 logger.log(logging.Level.INFO, 'This is an info message')
 ```
 
 You can also use the instance methods to target logs of a particular level:
-```
+```js
 logger.finest('this is the finest message')
 logger.finer('this is a finer message')
 logger.info('this is an info message')
@@ -120,24 +120,24 @@ logger.severe('this is a severe message')
 
 By default the logger's level is `Level.OFF`
 To change the level of the logger:
-```
+```js
 logger.setLevel(logging.Level.INFO)
 ```
 
 ### 3. Log Output:
 
 As opposed to ruby which by default sends logs to console in stdout, JS has the option to add a handler. The handler will be invoked for each message logged with this instance, or any of its descendants.
-```
+```js
 logger.addHandler(callback)
 ```
 
 JS has provided a method to add console handler to the given logger. The console handler will log all messages using the JavaScript Console API:
-```
+```js
 logging.addConsoleHandler(logger)
 ```
 
 Similarly, to add console handler to the root logger:
-```
+```js
 logging.installConsoleHandler()
 ```
 
@@ -146,7 +146,7 @@ logging.installConsoleHandler()
 JS gives the ability to define log preference for a remote WebDriver session. There are 5 common log types: `BROWSER`, `CLIENT`, `DRIVER`, `PERFORMANCE`, `SERVER`
 
 To set a log level for a particular log type:
-```
+```js
 let prefs = new logging.Preferences()
 prefs.setLevel(logging.Type.BROWSER, logging.Level.FINE)
 ```
--- a/website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
+++ b/website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
@@ -129,10 +129,10 @@ Different languages have different level to log information.
   {{< alert-content >}}{{< /alert-content >}}
   {{< /tab >}}
   {{< tab header="Ruby" >}}
-  Logs request and response at level: <code>:info</code>
+  Logs useful information at level: <code>:info</code>
   {{< /tab >}}
   {{< tab header="JavaScript" >}}
-  Logs request and response at level: <code>FINER</code>
+  Logs useful information at level: <code>FINER</code>
   {{< /tab >}}
   {{< tab header="Kotlin" >}}
   {{< alert-content >}}{{< /alert-content >}}
@@ -142,6 +142,27 @@ Different languages have different level to log information.
 **Debugging Details**
 
 The debug log level is used for information that may be needed for diagnosing issues and troubleshooting problems.
+
+{{< tabpane langEqualsHeader=true text=true >}}
+  {{< tab header="Java" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="Python" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="CSharp" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="Ruby" >}}
+  Logs debugging details at level: <code>:info</code> and <code>:debug</code>
+  {{< /tab >}}
+  {{< tab header="JavaScript" >}}
+  Logs debugging details at level: <code>FINER</code> and <code>FINEST</code>
+  {{< /tab >}}
+  {{< tab header="Kotlin" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+{{< /tabpane >}}
 
 ### 3. Log Output:
 Logs can be displayed on `stdout` or stored in a file.
--- a/website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
+++ b/website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
@@ -68,7 +68,7 @@ Logger level helps to filter out logs based on their severity.
 
 **Actionable Items**
 
-Things are logged as warnings if they are something the user needs to take action on. This is mostly used
+Things are logged as warnings if they are something the user needs to take action on. This is often used
 for deprecations. For various reasons, Selenium project does not follow standard Semantic Versioning practices.
 Our policy is to mark things as deprecated for 3 releases and then remove them.
 
@@ -129,10 +129,12 @@ Different languages have different level to log information.
   {{< alert-content >}}{{< /alert-content >}}
   {{< /tab >}}
   {{< tab header="Ruby" >}}
-  Logs useful information at level: <code>:info</code>
+  <p>Because the Ruby <code>Logger</code> class only has one "debug" level, Selenium is currently using the <code>:info</code> level as a general debug mode, and <code>:debug</code> as a lower level debug mode.
+  To bring things in line with other languages, we are considering <a href="https://github.com/SeleniumHQ/selenium/issues/11797">changing this behavior</a>.
+  For now, both info and warnings are handled at the default <code>:warn</code> level.</p>
   {{< /tab >}}
   {{< tab header="JavaScript" >}}
-  Logs useful information at level: <code>FINER</code>
+  Logs useful information at level: <code>INFO</code>
   {{< /tab >}}
   {{< tab header="Kotlin" >}}
   {{< alert-content >}}{{< /alert-content >}}
@@ -184,10 +186,10 @@ Logs can be displayed on `stdout` or stored in a file.
   Selenium::WebDriver.logger.output = '/path/to/selenium.log'</pre>
   <br>
   <p>Sample Output:</p>
-  <blockquote>
-  2023-02-08 12:31:23 INFO Selenium -> POST session/8b83ff54712d0a247937d045f8c8e171/url<br>
-  2023-02-08 12:31:23 INFO Selenium    >>> http://127.0.0.1:9515/session/8b83ff54712d0a247937d045f8c8e171/url | {"url":"https://www.selenium.dev/selenium/web/web-form.html"}<br>
-  2023-02-08 12:31:23 INFO Selenium <- {"value":null}</blockquote>
+  <pre>
+  2023-02-08 12:31:23 INFO Selenium -> POST session/8b83ff54712d0a247937d045f8c8e171/url
+  2023-02-08 12:31:23 INFO Selenium    >>> http://127.0.0.1:9515/session/8b83ff54712d0a247937d045f8c8e171/url | {"url":"https://www.selenium.dev/selenium/web/web-form.html"}
+  2023-02-08 12:31:23 INFO Selenium <- {"value":null}</pre>
   {{< /tab >}}
   {{< tab header="JavaScript" >}}
   <p>Send logs to console output:</p>
@@ -195,7 +197,7 @@ Logs can be displayed on `stdout` or stored in a file.
   logging.installConsoleHandler()</pre>
   <br>
   <p>Sample Output:</p>
-  <blockquote>
+  <pre>
   [2023-03-21T22:28:20Z] [FINER] [webdriver.http.Executor] >>> POST /session/0208994573cca3250a1066424c1b915c/url<br>
   [2023-03-21T22:28:22Z] [FINER] [webdriver.http.Executor] >>>
   POST /session/0208994573cca3250a1066424c1b915c/url HTTP/1.1<br>
@@ -208,7 +210,8 @@ Logs can be displayed on `stdout` or stored in a file.
   content-type: application/json; charset=utf-8
   cache-control: no-cache
 
-  {"value":null}</blockquote>
+  {"value":null}</pre>
+  <br>
   {{< alert-content >}}
   Store logs in a file.
   {{< /alert-content >}}
--- a/website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
+++ b/website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
@@ -21,16 +21,15 @@ of the program.
   {{< tab header="CSharp" >}}
   {{< alert-content >}}{{< /alert-content >}}
   {{< /tab >}}
-  {{< tab header="Ruby" >}}
-  <p>Ruby uses a custom implementation of the default `Logger` class with some interesting additional features.</p>
-  <pre>
-  logger = Selenium::WebDriver.logger</pre>
-  {{< /tab >}}
-  {{< tab header="JavaScript" >}}
-  <pre>
-  const logging = require('selenium-webdriver/lib/logging')
-  logger = logging.getLogger('webdriver')</pre>
-  {{< /tab >}}
+  {{% tab header="Ruby" %}}
+  Ruby uses a custom implementation of the default `Logger` class with some interesting additional features.
+    
+    logger = Selenium::WebDriver.logger
+  {{% /tab %}}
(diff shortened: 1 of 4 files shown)


titusfortner:
I think we should put this whole page in a tab rather than bullet points per language. What do you think?

titusfortner:
This is formatted incorrectly. We changed how tabs need to display code with the latest Docsy update, needs to be `text=true` instead of `code=false`

And can we move & update the `{{< alert-content >}}` lines to go into the tabs that don't have information?

Thanks!

TamsilAmani:
Thank you for the help regarding formatting @titusfortner. I have pushed the changes.

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,70 +9,165 @@ description: >
...
+{{< /tab >}}
+
+{{< tab header="JavaScript" >}}
+<p>JS has implemented logging APIs. It has also added LogManager to maintain a collection of loggers.</p>
+<p><b>1. Getting a logger can be done in two ways:</b></p>
+<pre>
+  const logging = require('selenium-webdriver/lib/logging')
+</pre>
+<br>
+<p>First method using manager</p>
```
titusfortner: Is the only reason to use the manager if you need more than one logger?

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,70 +9,165 @@ description: >
...
+</pre>
+<br>
+<p>Second method without manager</p>
+<pre>
+  logger = logging.getLogger('')
+</pre>
+<br>
+<p>Loggers class uses hierarchical, dot-delimited namespaces (e.g. "" > "webdriver" > "webdriver.logging"). For example, the following code will give a hierarchy of loggers:</p>
+<pre>
+  logger = mgr.getLogger('foo.bar.baz')
```
titusfortner: Can we use a real Selenium example instead of `foo.baz.bar` I'm still not sure I get what this is doing.

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,70 +9,165 @@ description: >
...
+</pre>
+<br>
+<p>"" (parent of) > "foo" (parent of) > "foo.bar" (parent of) > "foo.bar.baz" (=logger)</p>
+
+<p><b>2. Logger Level:</b></p>
+
+<p>The basic levels for JS loggers are: <i>`OFF`, `SEVERE`, `WARNING`, `INFO`, `DEBUG`, `FINE`, `FINER`, `FINEST`, `ALL`</i>.<br>
+To log messages at the given level pass the level in the <i>`log()`</i> method:</p>
```
titusfortner: How to log messages is important for writing Selenium code, but this page is intended to be for users who need to look at the logs coming from Selenium.

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,70 +9,165 @@ description: >
...
+</pre>
+<br>
+<p>By default the logger's level is <i>`Level.OFF`.</i> To change the level of the logger:</p>
+<pre>
+  logger.setLevel(logging.Level.INFO)
+</pre>
+<br>
+<p><b>3. Log Output:</b></p>
+
+<p>As opposed to ruby which by default sends logs to console in stdout, JS has the option to add a handler. The handler will be invoked for each message logged with this instance, or any of its descendants.</p>
```
titusfortner: So what is the default output for JS? Do you have to add a handler?

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,70 +9,165 @@ description: >
...
+<pre>
+  logging.addConsoleHandler(logger)
+</pre>
+<br>
+<p>Similarly, to add console handler to the root logger:</p>
+<pre>
+  logging.installConsoleHandler()
+</pre>
+<br>
+<p><b>4. Logging Preference (remote logging API):</b></p>
```
titusfortner: This is actually not maintained any longer, and shouldn't be included. :(

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,70 +9,165 @@ description: >
...
+</pre>
+<br>
+<p><b>3. Log Output:</b></p>
+
+<p>As opposed to ruby which by default sends logs to console in stdout, JS has the option to add a handler. The handler will be invoked for each message logged with this instance, or any of its descendants.</p>
+<pre>
+  logger.addHandler(callback)
+</pre>
+<br>
+<p>JS has provided a method to add console handler to the given logger. The console handler will log all messages using the JavaScript Console API:</p>
```
titusfortner: It is probably an issue of me not understanding JavaScript, but what is the difference between a handler and a console handler?

titusfortner (changes requested):
A few things I'm not clear on...

Also, can we get an example of what actual output looks like based on example code?

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
TamsilAmani: I looked further into LogManager. Though it helps us to store loggers it is not required in our case. We don't need to explicitly use a LogManager. Selenium maintains a LogManager when it starts. I have cleaned the approach in the new commit.

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
TamsilAmani: No need to use this as we are not creating a logger or using a manager. We are just getting the already created logger 'webdriver'.

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
TamsilAmani: Thank you for the clarity. This helped much. I was focussing on users' logging too. But now I have changed the approach to only target logs which Selenium generates.

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
TamsilAmani: There is no default output for JS. You have to use a method `installConsoleHandler()` to send the logs to the console.
Storing in a file is also not possible as of now.

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
TamsilAmani: Got it. Removed it in the latest commit.

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
TamsilAmani: A handler will take a callback function and it will be invoked when the event(log) occurs. Console handler is a special type of handler which will display the logs on console in a fixed format. 
https://github.com/SeleniumHQ/selenium/blob/trunk/javascript/node/selenium-webdriver/lib/logging.js#L526-L559

But for user's perspective to display selenium logs, we will use `installConsoleHandler()`.

TamsilAmani:
> A few things I'm not clear on...
> 
> Also, can we get an example of what actual output looks like based on example code?

Can you elaborate this a little please @titusfortner? Do you want to show outputs for different languages in their respective tabs along with a working sample code?

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
titusfortner: Wait, there is no file handler that can be used?

titusfortner:
> Do you want to show outputs for different languages in their respective tabs along with a working sample code

I meant like just a line showing actual output. Like this:

https://github.com/SeleniumHQ/seleniumhq.github.io/blob/trunk/website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md?plain=1#L44-L45

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,55 +9,105 @@ description: >
...
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="Python" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="CSharp" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="Ruby" >}}
+  <p>Ruby uses a custom implementation of the default `Logger` class with some interesting additional features.</p>
```
titusfortner: <p>Ruby uses a custom implementation of the default `Logger` class with some interesting additional features.</p>
```
logger = Selenium::WebDriver.logger
```

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,55 +9,105 @@ description: >
...
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="Python" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="CSharp" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="Ruby" >}}
+  <p>Levels are: <i>`:debug`, `:info`, `:warn`, `:error`, `:fatal`.</i> Default is <i>`:warn`</i>.</p>
```
titusfortner: These are code symbols, I wouldn't put them in italics

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,55 +9,105 @@ description: >
...
+  <p>To change the level of the logger:</p>
+  <pre>
+  Selenium::WebDriver.logger.level = :fatal</pre>
```
titusfortner:
```suggestion
  Selenium::WebDriver.logger.level = :info</pre>
```

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,55 +9,105 @@ description: >
...
+  <p>To change the level of the logger:</p>
+  <pre>
+  logger.setLevel(logging.Level.INFO)</pre>
+  {{< /tab >}}
+  {{< tab header="Kotlin" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+{{< /tabpane >}}
 
 **WARN**
```
titusfortner: Instead of the Ruby constants, let's make these headers normal English:
**Actionable Items**

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,55 +9,105 @@ description: >
...
 
 Warnings include everything we want users to be aware of by default. This is mostly used
```
titusfortner:
```suggestion
Things are logged as warnings if they are something the user needs to take action on. This is mostly used
```

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,55 +9,105 @@ description: >
...
+  {{< /tab >}}
+  {{< tab header="JavaScript" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="Kotlin" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+{{< /tabpane >}}
 
 **INFO**
```
titusfortner:
```suggestion
**Useful Information**
```

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -66,13 +116,63 @@ sent to and received from the driver or server. This is a great way to see what
 doing under the hood, and can be used to determine if it is Selenium code or driver code that
```
titusfortner: This is only true for Ruby. We should move all of this to debugging info section for Ruby.
"Useful Information" section should say something like: "This is the default level where Selenium logs things that users should be aware of but do not need to take actions on."

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -66,13 +116,63 @@ sent to and received from the driver or server. This is a great way to see what
...
+  {{< /tab >}}
+  {{< tab header="JavaScript" >}}
+  Logs request and response at level: <i>`FINER`</i>
+  {{< /tab >}}
+  {{< tab header="Kotlin" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+{{< /tabpane >}}
+
 **DEBUG**
```
titusfortner:
```suggestion
**Debugging Details**
```

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -66,13 +116,63 @@ sent to and received from the driver or server. This is a great way to see what
...
 
 This is less useful information where we log things about the servers and the sockets, and header information, etc.
```
titusfortner: Again these words only apply for Ruby, so can be moved to the Ruby only section. Don't worry about Ruby tab too much, put the right words that will apply to all logging and I'll add some things to the Ruby tab to clarify.

titusfortner (changes requested):
A few thoughts to make it more language-agnostic. I can update the Ruby tab things after the next pass. Thanks!

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
TamsilAmani: Though I have written a general statement wrt debug level, the question which arises is are we targeting the same `'debug'` level in all languages? Cause almost every lang has a debug level but the logs which they show may not be the same.

For example, you have already compared ruby and python [here](https://gist.github.com/titusfortner/ef2b4c4d3ebc32dd91e44d55c2d5d2d5). (https://github.com/SeleniumHQ/selenium/issues/10944#issuecomment-1423297676)

I have used the same [script](https://www.selenium.dev/documentation/webdriver/getting_started/first_script/#putting-everything-together) (JS version) and found that no logs are showing for DEBUG level.

What do you suggest in this case? @titusfortner

titusfortner: Yeah, the suggestion is to have the headline be "Debugging details" or similar and then describe which settings apply for that language. In Java & JS that's going to be `FINER` for most things, right? For Ruby, right now, it's both `:info` and `:debug` (and I created an issue to make them match here https://github.com/SeleniumHQ/selenium/issues/11797)

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
TamsilAmani: At present no API is there to store logs in a file like ruby does. But as discussed in TLC, I will implement the API for the same later.

TamsilAmani:
Please have a look @titusfortner. Let me know if any changes are to be made. This PR only adds JS and ruby sections. For other bindings I'll make separate PRs.

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,70 +9,211 @@ description: >
...
+  logger.setLevel(logging.Level.INFO)</pre>
+  {{< /tab >}}
+  {{< tab header="Kotlin" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+{{< /tabpane >}}
+
+**Actionable Items**
+
+Things are logged as warnings if they are something the user needs to take action on. This is mostly used
```
titusfortner:
```suggestion
Things are logged as warnings if they are something the user needs to take action on. This is often used
```

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,70 +9,211 @@ description: >
...
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="Python" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="CSharp" >}}
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="Ruby" >}}
+  Logs useful information at level: <code>:info</code>
```
titusfortner:
```suggestion
Because the Ruby `Logger` class only has one "debug" level, Selenium is currently using the `:info` level as a general debug mode, and `:debug` as a lower level debug mode.
To bring things in line with other languages, we are considering [changing this behavior]({{< ref "https://github.com/SeleniumHQ/selenium/issues/11797" >}}).
For now, both info and warnings are handled at the default `:warn` level.
```

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,70 +9,211 @@ description: >
...
+  {{< /tab >}}
+  {{< tab header="JavaScript" >}}
+  Logs useful information at level: <code>FINER</code>
```
titusfortner:
```suggestion
  Logs useful information at level: <code>INFO</code>
```

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,70 +9,211 @@ description: >
...
+  {{< alert-content >}}{{< /alert-content >}}
+  {{< /tab >}}
+  {{< tab header="Ruby" >}}
+  <p>By default, logs are sent to the console in `stdout`.</p>
+  <p>To store the logs in a file:</p>
+  <pre>
+  Selenium::WebDriver.logger.output = '/path/to/selenium.log'</pre>
+  <br>
+  <p>Sample Output:</p>
+  <blockquote>
```
titusfortner: This formatting is hard to read, I think we need to treat this as a codeblock instead of a blockquote so that it uses a mono format.

website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
```diff
@@ -9,70 +9,211 @@ description: >
...
+  2023-02-08 12:31:23 INFO Selenium    >>> http://127.0.0.1:9515/session/8b83ff54712d0a247937d045f8c8e171/url | {"url":"https://www.selenium.dev/selenium/web/web-form.html"}<br>
+  2023-02-08 12:31:23 INFO Selenium <- {"value":null}</blockquote>
+  {{< /tab >}}
+  {{< tab header="JavaScript" >}}
+  <p>Send logs to console output:</p>
+  <pre>
+  logging.installConsoleHandler()</pre>
+  <br>
+  <p>Sample Output:</p>
+  <blockquote>
```
titusfortner: same as above, codeblock not blockquote

titusfortner (changes requested):
Looks good! A couple minor tweaks.

TamsilAmani:
Changes pushed.

titusfortner:
Oh, and this probably should use markdown instead of html.

https://www.selenium.dev/documentation/about/style/#using-markdown-in-a-tab

need to use `{{% tab header="Java" %}}` instead of `{{< tab header="Java" >}}`

titusfortner (approved):

TamsilAmani:
Ah! Thank you for that link. I was just wondering how to use MD instead of HTML and was all over the internet for the solution. Will push the changes soon.

titusfortner:
lol, yeah, it isn't obvious. I only know because I was paying attention to those specific changes when they were made (we didn't used to be able to do text in the tabs of any kind).

diemol:
I guess we need to have the files in the other languages in order to merge this.

TamsilAmani:
@diemol So should I push the docs for Java and Python in this same branch? (Also I haven't explored how things are done in C# and Kotlin hence that might take time)

titusfortner:
No, he means the other translated md files. Just copy paste what you have over and our amazing translators will do the rest.

TamsilAmani:
@titusfortner The `alert-content` is not working with markdown style (%). So I have left the tabs which use alerts to still use html (<>). The markdown style can be done when these tabs are filled with their content.

@diemol I have added other language file.

diemol (approved):
Thank you, @TamsilAmani!

Issue #1298, opened 2023-01-27
[🐛 Bug]: Test Runners Section in Documention on 'Write your first Selenium script' just contains placeholder texts.


ashutoshdubey133:
### What happened?

The Test Runners section in the documentation just has the following comment
-  //Add instruction

I think we are missing actual test runner examples. If someone can guide me in the right direction, I can look into it and create a PR.

This is the link - 
[Link](https://www.selenium.dev/documentation/webdriver/getting_started/first_script/#test-runners)

![image](https://user-images.githubusercontent.com/54872020/215077755-c1aa420c-842b-452b-acee-99e52b356e83.png)

### What browsers and operating systems are you seeing the problem on?

Not a browser specific problem.

GavinHaydy:
maybe I can do it

Grs257:
Need sometime

rpallavisharma:
I believe this bug is closed, i see information in test runner section - https://www.selenium.dev/documentation/webdriver/getting_started/using_selenium/

PR #1324, opened 2023-03-01
update TestRunners for first_script.en.md

GavinHaydy:

### Description

 add document for firstScript  #1298 
### Types of changes

- [ ] Improved translation

### Checklist


website_and_docs/content/documentation/webdriver/getting_started/first_script.en.md
```diff
@@ -256,16 +256,38 @@ There are multiple options in each language, but here is what we are using in ou
...
-// Add instructions
+```shell
+# install Dependent library
+mvn install
+
+# run all case
+mvn test  
+
+#  run FirstScriptTest only   
+# Modify "**/*Test.java" to "**/FirstScriptTest" add Comment out "<include>**/*Example.java</include>" in pom.xml
```
diemol: Why?

examples/python/tests/getting_started/test_first_script.py
```diff
@@ -23,3 +23,6 @@ def test_eight_components():
     assert value == "Received!"
 
     driver.quit()
+
+
+test_eight_components()
```
diemol: This is not needed because we use `pytest` to run the examples. So ideally the instructions should mention `pytest`.

website_and_docs/content/documentation/webdriver/getting_started/first_script.en.md
diemol: Maybe it is better to tell them how we are running the examples?

diemol (changes requested):
Thank you, @the-ruffian!

I left some comments.

website_and_docs/content/documentation/webdriver/getting_started/first_script.en.md
GavinHaydy: > 

I misunderstood when I thought I only needed to run the FirstScriptTest file

GavinHaydy: > 也许最好告诉他们我们是如何运行示例的？
I'm going to modify it

examples/python/tests/getting_started/test_first_script.py
GavinHaydy: > 这不是必需的，因为我们用来`pytest`运行示例。所以理想情况下，说明书应该提到`pytest`。
Okay, I was negligent

diemol:
Actually I am going to let @titusfortner check it because he has a better understanding on how this should look like.

titusfortner:
I see that we're getting these to match what JS has, but if a user doesn't already know these things, reading this in the documentation is not going to help them. 

I think we actually need to create a separate page for this and create a full tutorial.

Probably the best thing to do in this section right now is to describe why to use a test framework, list options, and then link to what we've done in our examples (https://github.com/SeleniumHQ/seleniumhq.github.io/tree/trunk/examples/)

GavinHaydy:
> I see that we're getting these to match what JS has, but if a user doesn't already know these things, reading this in the documentation is not going to help them.
> 
> I think we actually need to create a separate page for this and create a full tutorial.
> 
> Probably the best thing to do in this section right now is to describe why to use a test framework, list options, and then link to what we've done in our examples (https://github.com/SeleniumHQ/seleniumhq.github.io/tree/trunk/examples/)

That's a good idea. I'll try to do it

GavinHaydy:
@titusfortner Maybe we can  create a “Install Unit test framework” between “Install Driver” and "Fires Script" to write the unit test framework and project configuration.

titusfortner:
Well, you don't need a test runner to use Selenium if you just want to write a quick script and execute it.

We're trying to support the idea that Selenium is used for more than testing, and that a test framework with a test runner is overkill unless attempting to scale up a testing solution.

"First script" is supposed to be the bare minimum requirements to get Selenium to execute.

But then we point out how testers need to scale up code for a successful testing strategy.

the-ruffian:
But in "Putting everything together", we used mocha. If want the minimum requirements,Maybe I need to write a complete and simple script in "Test Runners"

GavinHaydy:
I finished.
Please check it

PR #1355, opened 2023-04-06
allow designated TLC representative to the PLC

titusfortner:
### Description
Change governance document for the TLC to nominate a PLC member for a pre-designated term.

### Motivation and Context
This was discussed in our meeting at the Selenium Conference. There used to be significant overlap in membership between TLC and PLC. This is a good way to provide more crossover than we have now.


p0deje (approved):

diemol (approved):

AutomatedTester:
Since this removes the employer restriction on the TLC member in the PLC, do they have similar voting rights as PLC members or do they have reduced voting powers to prevent potential issues of 1 company trying to exert control over the project?

diemol:
> 

As far as I understood, yes. But the vote should be what the TLC thinks. Not what the designated person thinks. This means, if a vote is needed, the designated person should consult the rest of TLC members for their opinion. However, this is not written in the text. 

Again, this is what I have understood so far. @titusfortner was it like that?

diemol:
> > 
> 
> As far as I understood, yes. But the vote should be what the TLC thinks. Not what the designated person thinks. This means, if a vote is needed, the designated person should consult the rest of TLC members for their opinion. However, this is not written in the text.
> 
> Again, this is what I have understood so far. @titusfortner was it like that?

@AutomatedTester if this detail about the voting process is added to this PR, would you be fine with it? Is there something else you would like to check?

AutomatedTester:
Yes, if there is text saying that the TLC member on the PLC is just proxying the TLC vote then it's fine.

mmerrell (approved):

website_and_docs/content/project/governance/_index.html
```diff
@@ -428,12 +428,18 @@ <h3 id="project-leadership-committee-plc">Project Leadership Committee (PLC)</h3
...
+        Standard PLC seats are not time-limited. Only one person on the PLC may be affiliated with
         any given employer at a time. The PLC member count should be always
         odd so that no votes are ever tied, an ideal minimum number is 5. PLC members
         should also be active members of the community.
       </p>
 
+      <p class="pb-2">
+        Additionally, the TLC may conduct a vote to appoint one of its members to serve as an active PLC member.
+        This person is not be subject to the employer restriction above, but the appointment automatically expires 
+	at the end of six months.
```
AutomatedTester: We need to have something like 

> The TLC member on the PLC will use their vote to pass on the wishes of the TLC. This may require that the member hold a vote from TLC members to create consensus to be proxied up to the TLC.

diemol: https://github.com/SeleniumHQ/seleniumhq.github.io/commit/2dcc335fd6ab409891968a9a462cab3da19d6db6

manoj9788 (approved):
Looks good to me!Thanks for the language correction @AutomatedTester 
Thanks @titusfortner @diemol for actively pushing this!

diemol:
So far we have 3 👍 from the PLC, @wcmcgee and @maaretp are remaining.

image

testbot206:
I added a thumbs up but I don't know if it's in the right place 🙄

website_and_docs/content/project/governance/_index.html
```diff
@@ -428,12 +428,19 @@ <h3 id="project-leadership-committee-plc">Project Leadership Committee (PLC)</h3
...
 
       <p class="pb-2">
-        PLC seats are not time-limited. Only one person on the PLC may be affiliated with
+        Standard PLC seats are not time-limited. Only one person on the PLC may be affiliated with
         any given employer at a time. The PLC member count should be always
         odd so that no votes are ever tied, an ideal minimum number is 5. PLC members
         should also be active members of the community.
       </p>
 
+      <p class="pb-2">
```
maaretp, line 437: This language should be appended with information on how the voting in case of tie works. In many non-profits bylaws, the chair vote is considered to have winning emphasis on case of a tie, and if the TLC vote represents multiple people, it might make sense to either allow that or actively remove that.

maaretp (approved):

maaretp:
We have multiple options on how the 6th voting member through TLC in PLC could work: 

1. No voting right -- seems senseless
2. Single vote that in tie sets the winning vote -- would seem befitting for this project
3. Single vote where in tie a chairman's cote sets the winning vote -- should allocate a chair for PLC which we don't have and does not feel like it would be right.

diemol:
> We have multiple options on how the 6th voting member through TLC in PLC could work:
> 
>     1. No voting right -- seems senseless
> 
>     2. Single vote that in tie sets the winning vote -- would seem befitting for this project
> 
>     3. Single vote where in tie a chairman's cote sets the winning vote -- should allocate a chair for PLC which we don't have and does not feel like it would be right.

We haven't merged this because this is a very good question and we need to find an answer.

diemol:
After asking in the channel, it seems folks like option two for voting. 

image

website_and_docs/content/project/governance/_index.html
diemol: Addressed this in the recent commit.

PR #1358, opened 2023-04-13
update wait documentation

titusfortner:
### Description
* Removed the static code examples and replaced with references code in the examples directory
* Added a section on hard coded sleeps; not technically a Selenium "feature," so maybe we don't want this?
* Removed Java-specific comments
* Removed the "Fluent Wait" section because that term only applies to Java (the example for customizing the explicit wait for Java uses the `FluentWait` class)
* Moved Expected Conditions out of the Wait page and into the Support Features page
* All explicit wait examples use lambdas not conditions classes
* Hopefully the explanations are more clear and focuses on the important bits

### Motivation and Context
* Code examples were not working
* Descriptions were repetitive and overly technical

### DIFF
--- a/website_and_docs/content/documentation/webdriver/support_features/expected_conditions.en.md
+++ b/website_and_docs/content/documentation/webdriver/support_features/expected_conditions.en.md
@@ -1,0 +1,43 @@
+---
+title: "Waiting with Expected Conditions"
+linkTitle: "Expected Conditions"
+weight: 1
+description: >
+  These are classes used to describe what needs to be waited for.
+---
+
+Expected Conditions are used with [Explicit Waits](({{< ref "waits#explicit-waits" >}})).
+Instead of defining the block of code to be executed with a _lambda_, an expected
+conditions method can be created to represent common things that get waited on. Some
+methods take locators as arguments, others take elements as arguments.
+
+These methods can include conditions such as:
+
+* element exists
+* element is stale
+* element is visible
+* text is visible
+* title contains specified value
+
+{{< tabpane text=true langEqualsHeader=true >}}
+{{% tab header="Java" %}}
+[Expected Conditions Documentation](https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/support/ui/ExpectedConditions.html)
+{{< badge-code >}}
+{{% /tab %}}
+{{< tab header="Python" >}}
+[Expected Conditions Documentation](https://www.selenium.dev/selenium/docs/api/py/webdriver_support/selenium.webdriver.support.expected_conditions.html)
+{{< badge-code >}}
+{{< /tab >}}
+{{< tab header="CSharp" >}}
+.NET stopped supporting Expected Conditions in Selenium 4 to minimize maintenance hassle and redundancy.
+{{< /tab >}}
+{{< tab header="Ruby" >}}
+Ruby makes frequent use of blocks, procs and lambdas and does not need Expected Conditions classes
+{{< /tab >}}
+{{< tab header="JavaScript" >}}
+{{< badge-code >}}
+{{< /tab >}}
+{{< tab header="Kotlin" >}}
+{{< badge-code >}}
+{{< /tab >}}
+{{< /tabpane >}}
--- a/website_and_docs/content/documentation/webdriver/support_features/expected_conditions.en.md
+++ b/website_and_docs/content/documentation/webdriver/support_features/expected_conditions.en.md
@@ -6,7 +6,7 @@ description: >
   These are classes used to describe what needs to be waited for.
 ---
 
-Expected Conditions are used with [Explicit Waits](({{< ref "waits#explicit-waits" >}})).
+Expected Conditions are used with [Explicit Waits]({{< ref "../waits#explicit-waits" >}}).
 Instead of defining the block of code to be executed with a _lambda_, an expected
 conditions method can be created to represent common things that get waited on. Some
 methods take locators as arguments, others take elements as arguments.
--- a/website_and_docs/content/documentation/webdriver/waits.en.md
+++ b/website_and_docs/content/documentation/webdriver/waits.en.md
@@ -1,359 +1,95 @@
 ---
-title: "Waits"
+title: "Waiting Strategies"
 linkTitle: "Waits"
 weight: 6
 aliases: ["/documentation/en/webdriver/waits/"]
 ---
 
-WebDriver can generally be said to have a blocking API.
-Because it is an out-of-process library that
-_instructs_ the browser what to do,
-and because the web platform has an intrinsically asynchronous nature,
-WebDriver does not track the active, real-time state of the DOM.
-This comes with some challenges that we will discuss here.
+Perhaps the most common challenge for browser automation is ensuring
+that the web application is in a state to execute a particular
+Selenium command as desired. the processes often end up in
+a _race condition_ where sometimes the browser gets into the right
+state first (things work as intended) and sometimes the Selenium code
+executes first (things do not work as intended). This is the
+primary cause of _flaky tests_.
 
-From experience,
-most intermittent issues that arise from use of Selenium and WebDriver
-are connected to _race conditions_ that occur between
-the browser and the user's instructions.
-An example could be that the user instructs the browser to navigate to a page,
-then gets a **no such element** error
-when trying to find an element.
+All navigation commands wait for a specific `readyState` value
+based on the [page load strategy]({{< ref "drivers/options#pageloadstrategy" >}}) (the 
+default is `"complete"`) before the driver returns control to the code.
+The `readyState` only concerns itself with loading assets defined in the HTML, 
+but loaded JavaScript assets often result in changes to the site,
+and elements you need to interact with may not yet be on the page
+when the code is ready to execute the next Selenium command.
 
-Consider the following document:
+Similarly, in a lot of single page applications, elements get dynamically
+added to a page or change visibility based on a click. For example, the box
+element takes a second to show up after clicking on the "adder" button, so
+even trying to locate that button will error without some form of synchronization: 
 
-```html
-<!doctype html>
-<meta charset=utf-8>
-<title>Race Condition Example</title>
-
-<script>
-  var initialised = false;
-  window.addEventListener("load", function() {
-    var newElement = document.createElement("p");
-    newElement.textContent = "Hello from JavaScript!";
-    document.body.appendChild(newElement);
-    initialised = true;
-  });
-</script>
-```
-
-The WebDriver instructions might look innocent enough:
-
-{{< tabpane langEqualsHeader=true >}}
+{{< tabpane text=true langEqualsHeader=true >}}
   {{< tab header="Java" >}}
-driver.get("file:///race_condition.html");
-WebElement element = driver.findElement(By.tagName("p"));
-assertEquals(element.getText(), "Hello from JavaScript!");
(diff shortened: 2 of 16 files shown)


harsha509 (dismissed):
LGTM!

website_and_docs/content/documentation/webdriver/waits.en.md
```diff
@@ -1,495 +1,206 @@
...
-  });
-</script>
-```
-
-The WebDriver instructions might look innocent enough:
-
-{{< tabpane langEqualsHeader=true >}}
+Perhaps the most common challenge for browser automation is ensuring
+that the web application is in a state to execute a particular
+Selenium command as desired. the processes often end up in
```
iamalittletester: 'The' processes -> new phrase should start with upper case

website_and_docs/content/documentation/webdriver/waits.en.md
```diff
@@ -1,495 +1,206 @@
...
+a _race condition_ where sometimes the browser gets into the right
+state first (things work as intended) and sometimes the Selenium code
+executes first (things do not work as intended). This is the
+primary cause of _flaky tests_.
```
iamalittletester: Let's say it's 'one of the causes'. Others could be the environment, etc.

website_and_docs/content/documentation/webdriver/waits.en.md
```diff
@@ -1,495 +1,206 @@
...
+There are several synchronization strategies that can be used to properly wait for
+the application to be in the state you need it to be for the next Selenium command.
 
-Console.WriteLine(firstResult.Text);
-  {{< /tab >}}
-  {{< tab header="Ruby" >}}
-require 'selenium-webdriver'
-driver = Selenium::WebDriver.for :firefox
-wait = Selenium::WebDriver::Wait.new(:timeout => 10)
+## Hard-coded sleeps
```
iamalittletester: Either remove this section or keep it but mark the fact that this is not a Selenium specific functionality.

website_and_docs/content/documentation/webdriver/waits.en.md
```diff
@@ -1,495 +1,206 @@
...
-Because the wait will swallow _no such element_ errors
-that are raised when the element is not found,
-the condition will retry until the element is found.
-Then it will take the return value, a _WebElement_,
-and pass it back through to our script.
+This is a global setting that applies to every element location call for the entire session.
+The default value is `0`, which means that if the element is not found, it will
+immediately return an error. If an implicit wait is set, the driver will wait for the 
+duration of the provided value before returning the error. Note that as soon as the 
+element is located, the driver will return the value and your code may continue, so a larger
```
iamalittletester: "will return the element" i think? Or at least not the value cause i think the value refers to the timeout value?

iamalittletester: "continue executing"

website_and_docs/content/documentation/webdriver/waits.en.md
```diff
@@ -1,495 +1,206 @@
...
+implicit wait value won't necessarily increase the time of your session.
```
iamalittletester: i would rephrase to say something like "increase the duration of the test execution"

website_and_docs/content/documentation/webdriver/waits.en.md
```diff
@@ -1,495 +1,206 @@
...
+added to a page or change visibility based on a click. For example, the box
+element takes a second to show up after clicking on the "adder" button, so
+even trying to locate that button will error without some form of synchronization: 
+
+{{< tabpane text=true langEqualsHeader=true >}}
   {{< tab header="Java" >}}
-driver.get("file:///race_condition.html");
-WebElement element = driver.findElement(By.tagName("p"));
-assertEquals(element.getText(), "Hello from JavaScript!");
+{{< gh-codeblock path="examples/java/src/test/java/dev/selenium/waits/WaitsTest.java#L23-L28" >}}
```
iamalittletester: I would rewrite the test without the 'assert for exception', since the purpose of the test is not to make sure the assertion is thrown but for the element interaction to be possible. Below in a section there is a reference that says this code would not fail using a wait, but it's the other way around.

website_and_docs/content/documentation/webdriver/waits.en.md
```diff
@@ -1,495 +1,206 @@
...
-driver.implicitly_wait(10)
-driver.get("http://somedomain/url_that_delays_loading")
-my_dynamic_element = driver.find_element(By.ID, "myDynamicElement")
-  {{< /tab >}}
+## Explicit waits
+
+_Explicit waits_ are loops you add to your code that poll the application 
+for a specific condition to evaluate as true before it exits the loop and
+continues to the next command in your code. If the condition is not met before a designated timeout value, 
+the code will give a timeout error.
```
iamalittletester: In Java it's an Exception. Not sure of other languages but if they all throw Exceptions i think that would be a better term. This is valid across this page (in multiple locations).

website_and_docs/content/documentation/webdriver/waits.en.md
```diff
@@ -1,495 +1,206 @@
...
+
+There are many ways for the application not to be in the desired state,
+so explicit waits are a great choice to specify exactly what is desired
+in each place they are needed.
```
iamalittletester: I think this could be rephrased a bit.

website_and_docs/content/documentation/webdriver/waits.en.md
```diff
@@ -1,495 +1,206 @@
...
+the code will give a timeout error.
+
+There are many ways for the application not to be in the desired state,
+so explicit waits are a great choice to specify exactly what is desired
+in each place they are needed. 
+For example, an element must be both present in the DOM and 
+[displayed](({{< ref "elements/information/#is-displayed" >}})) on the page
+in order for Selenium to interact with it. 
+
+In this example, clicking the
```
iamalittletester: Maybe "in the below example"

website_and_docs/content/documentation/webdriver/waits.en.md
```diff
@@ -1,495 +1,206 @@
...
+{{< badge-code >}}
   {{< /tab >}}
 {{< /tabpane >}}
 
-## FluentWait
+### Customization
 
-FluentWait instance defines the maximum amount of time to wait for a condition,
-as well as the frequency with which to check the condition.
+The Wait class can be created with various parameters that will change how the conditions are evaluated.
```
iamalittletester: instantiated? instead of created

ollie-iterators:
If there are more areas that need their static code to be changed to references to the examples directory, I think there should be an issue added about the areas that need this to happen.

titusfortner:
@ollie-iterators if we merge the PR I just linked we can tag the static code with `{{ badge-examples }}` to indicate that it needs to be moved over.

I'm planning to record some screen captures next week walking through each step of contributing to the documentation to better document what we're asking people to do to help.

ollie-iterators:
> @ollie-iterators if we merge the PR I just linked we can tag the static code with `{{ badge-examples }}` to indicate that it needs to be moved over.
> 
> I'm planning to record some screen captures next week walking through each step of contributing to the documentation to better document what we're asking people to do to help.

Great.

website_and_docs/content/documentation/webdriver/waits.en.md
titusfortner: Good call, I updated the paragraph before and deleted this one. No need to have code examples.

website_and_docs/content/documentation/webdriver/waits.en.md
titusfortner: how about: "the driver will return the element reference"

website_and_docs/content/documentation/webdriver/waits.en.md
titusfortner: I'm trying to avoid use the word "session" instead of "test" to make it more "function-agnostic," but I think duration makes more sense than time.

website_and_docs/content/documentation/webdriver/waits.en.md
titusfortner: But you make a good point, I'm not writing a test suite showing that it fails when it is supposed to and passes when it is supposed to. Let me delete this code example and describe the scenario instead.

website_and_docs/content/documentation/webdriver/waits.en.md
titusfortner: Yeah, only Java & .NET make that distinction; plus it seems like the less technical term for describing what is happening.

website_and_docs/content/documentation/webdriver/waits.en.md
titusfortner: Good call:
> explicit waits are a great choice to specify the exact condition to wait for
in each place it is needed

titusfortner:
Oof, this totally fell off my radar; I was going to do a quick update to get these changes out there, and of course... the structure is still confusing...

titusfortner:
Ok, removed a bunch of repetitive things in explicit waits section, hopefully easier to read now.

iamalittletester (approved):

website_and_docs/content/documentation/webdriver/waits.en.md
```diff
@@ -1,495 +1,159 @@
...
+default value to wait for is `"complete"`) before the driver returns control to the code.
+The `readyState` only concerns itself with loading assets defined in the HTML, 
+but loaded JavaScript assets often result in changes to the site,
+and elements that need to be interacted with may not yet be on the page
+when the code is ready to execute the next Selenium command.
+
+Similarly, in a lot of single page applications, elements get dynamically
+added to a page or change visibility based on a click.
+An element must be both present and
+[displayed](({{< ref "elements/information/#is-displayed" >}})) on the page
```
harsha509: looks like an `( )` broke the link in here!

website_and_docs/content/documentation/webdriver/waits.en.md
```diff
@@ -1,495 +1,159 @@
...
+can fail when it doesn't sleep long enough. Alternately, if the value is set too high
+and a sleep statement is added in every place it is needed, the duration of
+the session can become prohibitive.
+
+Selenium provides two different mechanisms for synchronization that are better.
+
+
+## Implicit waits
+Selenium has a built-in way to automatically wait for elements called an _implicit wait_.
+An implicit wait value can be set either with the [timeouts](({{< ref "drivers/options#timeouts" >}}))
```
harsha509: here too, the link is broken due to extra `( )`

website_and_docs/content/documentation/webdriver/support_features/expected_conditions.en.md
```diff
@@ -0,0 +1,43 @@
+---
+title: "Waiting with Expected Conditions"
+linkTitle: "Expected Conditions"
+weight: 1
+description: >
+  These are classes used to describe what needs to be waited for.
+---
+
+Expected Conditions are used with [Explicit Waits](({{< ref "waits#explicit-waits" >}})).
```
harsha509: extra `()` broke the link!

harsha509 (changes requested):

titusfortner:
odd there were a couple other places with double parens that seemed to work. Oh, I bet it was the combination double parens and the anchor reference at the end of the link...
Fixed all of them anyway

harsha509 (approved):
Thank you @titusfortner 🎉

PR #1382, opened 2023-05-09
Update Logging Documentation

titusfortner:
### Description
Reformatted/reorganized the page a bit to fit markdown structure better
Added details about how to do logging in Ruby, Python & Java referencing example code

### Motivation and Context
I want a place to point to when someone has a failure and I want more information about what might be going on. Should be especially useful in latest versions of Selenium with Selenium Manager logging added.

I... was going to add assertions by parsing the log file, but I ran out of energy. Also, we can't run the PyTest example because the easiest way to get the logs is actually to make the test fail. ☹️ 

@TamsilAmani would also appreciate if you have a chance to look at this.
@mdmintz let me know if there's a better way to show people what to do with the Python stuff.

### DIFF
--- a/website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
+++ b/website_and_docs/content/documentation/webdriver/troubleshooting/logging.en.md
@@ -6,101 +6,158 @@ description: >
   Getting information about Selenium execution.
 ---
 
-Each language adopts a distinctly different approach to logging information about the activity
-of the program.
+Turning on logging is a valuable way to get extra information that might help you determine
+why you might be having a problem.
 
-### 1. Getting a logger
+## Getting a logger
 
 {{< tabpane langEqualsHeader=true text=true >}}
-  {{< tab header="Java" >}}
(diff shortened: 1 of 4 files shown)


mdmintz:
> ``Also, we can't run the PyTest example because the easiest way to get the logs is actually to make the test fail.``

With ``pytest``, you can make a test "fail" without actually causing the CI to fail.

To do that, use ``@pytest.mark.xfail`` above a test that fails.

More info: https://docs.pytest.org/en/7.1.x/how-to/skipping.html

diemol (commented):
Tests are not compiling.

titusfortner:
Hmm, it wasn't compiling because it was trying to load code from trunk, I thought it was supposed to get it from PR branch?

I merged the code to trunk so we can see this.

titusfortner:
@mdmintz I realized that the test doesn't even need to show output to console (so no need to fail it). I'm sending it to the log and can validate the file. Thanks for the help, though.

diemol (approved):

PR #1384, opened 2023-05-10
Java use Base Test class where possible

titusfortner:
### Description
This uses the base test class to store the driver variable definition, and quit the driver.
I updated all places in the docs that referenced the affected code.
Redirects the Firefox log output to dev/null

### Motivation and Context
The reason I started down this path was to limit how much was getting logged to console by Firefox. And I figured it's better to use BaseTest where it makes sense to do so in order to be able to set default behavior in the future.


diemol (approved):
Thank you!

PR #1385, opened 2023-05-10
Update install drivers documentation

titusfortner:
### Description
* Updates details about Selenium Manager
* List the 3rd party managers instead of showing code
* Remove all code references to 3rd party managers

### Motivation and Context
We are currently logging as `INFO` when we are using Selenium Manager. I'd like to be able to add a link somewhere to explain it. I was going to create a separate page, but decided that a shorter section describing at a high level how it is used with the bindings made the most sense.

Most of this PR is removing all code references to the 3rd party managers. This lists them with links instead of showing example code, and puts into perspective when one might want to use them.


titusfortner:
There we go, now it is deploying 🤷‍♂️  
https://deploy-preview-1385--selenium-dev.netlify.app/documentation/webdriver/getting_started/install_drivers/

diemol:
Seems python is failing?

diemol (approved):

PR #1392, opened 2023-05-27
Add examples for using Service classes

titusfortner:
### Description
Examples in 4 languages of the common features of the  Service classes: Location, port & log output

### Motivation and Context
This is mostly interesting to highlight what works and what doesn't for driver logs in the various languages.

(The next step is fixing the code so it is the same in all languages)

https://deploy-preview-1392--selenium-dev.netlify.app/documentation/webdriver/drivers/service/


titusfortner:
I pushed example code to trunk so that the preview will build.

diemol (approved):

PR #1412, opened 2023-06-26
All the Service class updates

titusfortner:
### Description
Adding a bunch of stuff to show off how Service Classes currently work.
I updated a bunch of Options stuff while I was at it.

Note: I'm doing the PR just to make sure all the code examples work. ~I still need to match up the lines of code with what is in the markdown files~; I've updated the English pages with the correct code lines; ~I still need to update the other translations~. I'll move it out of draft when I do that.

### Motivation and Context
Another step in my push to get logging in Selenium make sense across all language and all browsers and completely documented.

I added some Python example code that should work once we release 4.11
Still a bunch of stuff that needs to be implemented still

### DIFF
--- a/examples/dotnet/SeleniumDocs/Browsers/ChromeTest.cs
+++ b/examples/dotnet/SeleniumDocs/Browsers/ChromeTest.cs
@@ -1,39 +1,164 @@
 using System;
 using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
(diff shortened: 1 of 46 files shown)


harsha509:
I believe deploy preview will be a bit confusing, but pr looks good to me!

harsha509 (approved):

diemol (approved):

PR #1456, opened 2023-08-15
Update _index.html

asolntsev:
Add Selenide maintainer name

### Types of changes
- [ ] Improved translation

### Checklist

### DIFF
--- a/website_and_docs/content/ecosystem/_index.html
+++ b/website_and_docs/content/ecosystem/_index.html
@@ -397,7 +397,7 @@ aliases:
             </p>
           </th>
           <td>Java</td>
-          <td>Selenide</td>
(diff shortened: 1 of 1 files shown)


titusfortner:
Yes, thanks for the fix.

titusfortner:
Oh, I didn't add the [deploy site] to the merge 🤦 It'll get updated next time someone else does it the right way. Sorry.

asolntsev:
Shouldn't we simplify this process?
I think site changes should be deployed automatically on every commit. It's quite cheap and fast operation, right?

titusfortner:
Not every commit needs a deploy, though. Not sure why it was set up this way, though. Site was deployed when I added the minutes from the last meeting.

PR #1464, opened 2023-09-01
blog post for Selenium 4.12

titusfortner:
* Waiting for updates to Java & .NET
* Need to couble check links in preview


titusfortner:
I thought we were doing the Java update same day, since we aren't, I'm going to publish this rather than wait.

PR #1467, opened 2023-09-05
[dotnet] Simplify and add examples for bidi network intercepring

nvborisenko:
### Description
Provide examples to intercept network in .NET

### Motivation and Context
Fixing lack of docs

### Types of changes

- [ ] Improved translation

### Checklist

- [ ] I have read the [**contributing**](https://selenium.dev/documentation/en/contributing/) document.


diemol:
We need to check that failing test...

nvborisenko:
I don't mind.

titusfortner:
We were getting the known location of the driver from PATH, but we can't do that any more. The goal is to be able to show that it will use a driver passed in. Maybe we download a different driver to cache and pass it in?

nvborisenko:
I reproduced it locally:
```csharp
var service = ChromeDriverService.CreateDefaultService();
service.DriverServicePath = "C:\\Temp";

driver = new ChromeDriver(service);
```

And getting:
```
System.ArgumentNullException : Value cannot be null. (Parameter 'path2')
Path.Combine(String path1, String path2)
DriverService.Start() line 266
```

nvborisenko:
Do you think changes in this PR caused errors? I believe no. I am not planning to fix failed tests in scope of this PR.

titusfortner (changes requested):
Please replace all the code in the md files with:
```
{{< gh-codeblock path="examples/dotnet/SeleniumDocs/ChromeDevTools/NetworkInterceptor.cs#L13-L24" >}}
```

Because all the other tabs are using automatic code formatting, we need to put `text=true` in the tabpane and `code=true` in the CSharp tab.

I need to (soon) go through and update all the marking so that we know what needs to be fixed.

nvborisenko:
Interesting, I created PR to add examples for BiDi network intercepting. BUT .net binding doesn't implement any BiDi protocol. @titusfortner  Do you still think is it valid example?

jimevans:
We need to make a distinction here when we use the term "BiDi" in Selenium. There are features that are bidirectional in nature, and can be referred to as "BiDi" features. Network interception is one of these types of features. Note very carefully that use of the term "BiDi" in this context has no bearing whatsoever on what protocol is used to implement that feature. Ideally, it will one day be implemented using the WebDriver BiDi protocol; currently it's still a "BiDi" feature, but implemented using CDP. So, in this context, the .NET bindings most certainly _do_ implement a "BiDi protocol," it's just CDP, not WebDriver BiDi.

nvborisenko:
Thank you @jimevans , then show must go on.

nvborisenko:
> 



> Please replace all the code in the md files with:
> 
> ```
> {{< gh-codeblock path="examples/dotnet/SeleniumDocs/ChromeDevTools/NetworkInterceptor.cs#L13-L24" >}}
> ```
> 
> Because all the other tabs are using automatic code formatting, we need to put `text=true` in the tabpane and `code=true` in the CSharp tab.
> 
> I need to (soon) go through and update all the marking so that we know what needs to be fixed.

But why java examples are not done in this way? Java examples are "hardcoded", I would like to do the same. For instance I don't want to include `using` statements in the examples just to achieve more simplified way how it will look like for end user.

And, what if I change C# source code like adding new lines in the code, will it update examples accordingly?

titusfortner:
Because the Java examples are wrong. 
I just went through to try to make this process more transparent and prep things for people to make it easier: #1473 

When lines of code are updated in the examples, we need to make sure that the md files all get updated as necessary.
Yes, it's a pain, but it is really nice when everything is working (like everything in Selenium project)

diemol:
Also, this guarantees that code examples are working.

titusfortner:
We need to keep text=true on those tabs. 

Docsy had it right in their previous version when they had code be opt-in instead of text as opt-in. Without text=true, the website treats whatever you write as code

titusfortner:
I've been futzing with the docs and moving things around. I just moved the examples to the "correct" location (the test file locations are supposed to match the page locations, but the BiDi/CDP examples haven't been following that convention.

Since you created a 3rd test, I added a new section in docs to include a reference to it. Hmm, I think I misspelled something, it isn't building, will investigate.

titusfortner:
Looks good: https://deploy-preview-1467--selenium-dev.netlify.app/documentation/webdriver/bidirectional/bidi_api/

titusfortner:
Thanks @nvborisenko

Issue #1333, opened 2023-03-11
[🐛 Bug]: Python - JavaScript alerts, prompts and confirmations


paulotcj:
### What happened?

In the documentation, the first example for handling alerts in Python at the address: https://www.selenium.dev/documentation/webdriver/interactions/alerts/
 is the following:

`driver.find_element(By.LINK_TEXT, "See an example alert").click()

alert = wait.until(expected_conditions.alert_is_present())

text = alert.text
alert.accept()`

However, this code will fail.
The probable solution as the example below hints, should be:
`driver.find_element(By.LINK_TEXT, "See an example alert").click()

WebDriverWait(driver, 10).until(expected_conditions.alert_is_present())

alert = driver.switch_to.alert

alert.accept()`

My apologies if I am missing something here.

### What browsers and operating systems are you seeing the problem on?

Any

titusfortner:
Hmm, I think the expected condition does switch to the alert?

At any rate, we are trying to update all code examples to actual code in the examples directory of this repo so that we can ensure things are properly passing. (https://www.selenium.dev/documentation/about/style/#reference-github-examples)

Is there any chance you'd be interested in writing that code in:
https://github.com/SeleniumHQ/seleniumhq.github.io/tree/trunk/examples/python/tests/interactions/test_alerts.py so that we can see it passing? We want all code examples using code in selenium.dev, so alerts tests should use: https://www.selenium.dev/selenium/web/alerts.html

eaccmk:
Hi @paulotcj , @titusfortner ,
This issue will mostly be covered by https://github.com/SeleniumHQ/seleniumhq.github.io/pull/1470

shbenzer:
Can be closed. Current python code example is as follows:

    element = driver.find_element(By.LINK_TEXT, "See an example alert")
    element.click()
    wait = WebDriverWait(driver, timeout=2)
    alert = wait.until(lambda d : d.switch_to.alert)
    text = alert.text
    alert.accept()

diemol:
Thank you, @shbenzer!

PR #1470, opened 2023-09-09
Fix for #1333 | Python alert automation

eaccmk:
### Description
This PR fixes issue #1333 

### Motivation and Context
Bug fix and contribution to Selenium community 🎉 

### Types of changes
- [ ] Improved translation
- [x] Bug Fix

### Checklist


titusfortner:
First, thanks for the PR.

The intention of the code here is not to thoroughly test a page, but to provide examples that can be displayed here: https://www.selenium.dev/documentation/webdriver/interactions/alerts/#alerts

So we only need 3 working examples that we can link to, and they should be complete examples instead of abstracting out the common pieces. The point is to look at the complete method.

Finally, I'm confused about the sleeps. We shouldn't need to hard code sleeps. Is the Python Selenium code not doing what we expect it to?

eaccmk:
> First, thanks for the PR.
> 
> The intention of the code here is not to thoroughly test a page, but to provide examples that can be displayed here: https://www.selenium.dev/documentation/webdriver/interactions/alerts/#alerts
> 
> So we only need 3 working examples that we can link to, and they should be complete examples instead of abstracting out the common pieces. The point is to look at the complete method.
> 
> Finally, I'm confused about the sleeps. We shouldn't need to hard code sleeps. Is the Python Selenium code not doing what we expect it to?

Sure, I can update the PR.

How about 3 examples, like this one  ( I used this [Alert example](https://www.selenium.dev/documentation/webdriver/interactions/alerts/#alerts) )

```python
# Click the link to activate the alert
driver.find_element(By.ID, "alert").click()

# Store the alert in a variable
alert = driver.switch_to.alert

# Store the alert text in a variable
text = alert.text

#assert alert text
assert alert.text == "cheese"

# Press the OK button
alert.accept()
```

On
> We shouldn't need to hard code sleeps. 

It was the Alert pop up from browser that was taking time when I was hitting all the alert links. If I am putting 3 basic examples, hardcoded sleep is not required and works fine.

eaccmk:
Done ✅  
@titusfortner, ready for review.

titusfortner:
That test gets fixed by #1472

titusfortner:
But also, we want to run these with pytest in our CI, so we want to put these into their own test methods.

titusfortner:
* Please remove the comments, we want the comments to be in the docs (so they can be translated) not in the code.
* The convention we're using is to have 3 separate tests to match these 3: https://www.selenium.dev/documentation/webdriver/interactions/alerts/
* We want to then take the code you've written and reference it in the alerts md file. We want to remove all of this: 
 https://github.com/SeleniumHQ/seleniumhq.github.io/blob/trunk/website_and_docs/content/documentation/webdriver/interactions/alerts.en.md?plain=1#L43-L53 

and replace it with this:
```
{{< gh-codeblock path="examples/python/tests/interactions/test_alerts.py#LXX-LYY" >}}
```
 
* Finally we're looking to copy that block of code in the en file to the other translation markdown files as well.

If you want to discuss or collaborate, come find us in the #selneium-docs channel in the chat room: https://www.selenium.dev/support/#ChatRoom

diemol:
This went stale, closing.

PR #1472, opened 2023-09-11
Add examples to set binary

titusfortner:
### Description
* Fixes Java test failures (except the failures from Selenium bug fixed with https://github.com/SeleniumHQ/selenium/commit/73da1e8bcad94eb147c195a57a0daea5ee638434)
* Installs Chrome, Edge and Firefox on all the github actions
* Puts location of each binary into environment variables
* Calls set binary for each of those in all the browsers for Ruby, Python, Java & .NET
* At the point I had to update all the lines of code for all examples in all browser files, I did some reformatting of the files

### Motivation and Context
* We sorely need the examples for people who are struggling with Selenium Manager related things
* We can (should) replace these Github Actions steps with commands from Selenium Manager eventually


.github/workflows/dotnet-examples.yml
```diff
@@ -30,6 +30,33 @@ jobs:
...
+      if: matrix.os == 'windows-latest'
+      run: |
+        echo "CHROME_BIN=${{ steps.setup-chrome.outputs.chrome-path }}" >> $env:GITHUB_ENV
+        echo "EDGE_BIN=${{ steps.setup-edge.outputs.edge-path }}" >> $env:GITHUB_ENV
+    - name: Set ENV Other
+      if: matrix.os != 'windows-latest'
+      run: |
+        echo "CHROME_BIN=${{ steps.setup-chrome.outputs.chrome-path }}" >> "$GITHUB_ENV"
+        echo "EDGE_BIN=${{ steps.setup-edge.outputs.edge-path }}" >> "$GITHUB_ENV"
+        echo "FF_BIN=${{ steps.setup-firefox.outputs.firefox-path }}" >> "$GITHUB_ENV"
```
diemol, line 59: Is there a way to aggregate this in an action to avoid the same setup in all workflows?

diemol (commented):
Didn't check the failing tests but the rest looks fine.

titusfortner:
The Java failures are the `logsToConsole` tests related to a bug that I fixed in trunk.

I'm rerunning the .NET Mac tests; they've passed several times before the PR https://github.com/SeleniumHQ/seleniumhq.github.io/actions/workflows/dotnet-examples.yml

.github/workflows/dotnet-examples.yml
titusfortner: probably, but we're going to want to re-implement all of this with Selenium Manager. What do you think @bonigarcia ? Can we do this once we put Selenium Manager in `~/.cache`? We'd just need to get the browser location output and put it in an ENV.  Actually, would be *really interesting to create and distribute our own GitHub actions workflow. These repos are not being actively maintained - https://github.com/browser-actions/

PR #1479, opened 2023-09-15
Create new Getting Started page for Running Tests

titusfortner:
### Description
* Made First Script be standalone executable instead of part of a test runner
* Created a "Running Tests" page to put details about IDEs and Test Runners (essentially how to use Selenium once you figure out the mechanics)
* Moved existing Test Runner content from the bottom of first script page
* Hello Selenium in .NET had to be moved out of "Selenium Docs" because .NET doesn't like more than one `Main()` method in the same project

### Motivation and Context
* It was suggested that First Script should not be in the context of a test runner, which makes sense
* People are looking to add code for Test Runners, and the way it is currently structured feels weird.


rpallavisharma:
Hi @titusfortner I think this would be appropriate step forward. Reading Test Runner page/section we might have to say it in manner that one of reasons people use Selenium is for web automation. then for each language the different test runner available and then do we give examples with each of them? maybe we could or maybe not. At the same time we may decide upon that the code examples we are giving if uses test runner which would we be if there is any need to do this(not sure). 

Also it would be good to add information other than testing how selenium is being used- that i think is our (strength maybe)

Let me know if I can do something here.

website_and_docs/content/documentation/webdriver/getting_started/running_tests.en.md
```diff
@@ -0,0 +1,212 @@
...
+{{< /tab >}}
+{{< tab header="Kotlin" >}}
+{{< badge-code >}}
+{{< /tab >}}
+{{< /tabpane >}}
+
+### Executing
+
+{{< tabpane text=true langEqualsHeader=true >}}
+{{< tab header="Java" >}}
```
harsha509: {{% tab header="Java" %}}

website_and_docs/content/documentation/webdriver/getting_started/running_tests.en.md
```diff
@@ -0,0 +1,212 @@
...
+{{% tab header="Python" %}}
+{{< badge-code >}}
+{{% /tab %}}
+{{< tab header="CSharp" >}}
+{{< badge-code >}}
+{{< /tab >}}
+{{< tab header="Ruby" >}}
+{{< badge-code >}}
+{{< /tab >}}
+{{< tab header="JavaScript" >}}
```
harsha509: {{% tab header="JavaScript" %}}

website_and_docs/content/documentation/webdriver/getting_started/running_tests.en.md
```diff
@@ -0,0 +1,212 @@
...
+```shell
+mocha runningTests.spec.js
+```
+{{< /tab >}}
+{{< tab header="Kotlin" >}}
```
harsha509: {{% tab header="Kotlin" %}}

harsha509 (changes requested):

titusfortner:
Good points.  Maybe instead of "Running Tests"It should be "Using Selenium." Then we can add a section on Use Cases? What else other than "automate repetitive tasks" would you suggest including?

I don't want to give examples with each test runner. I want to give links to all the test runners, and then show how this project is using the test runners we chose with references to actual code. To an extent, Selenium documentation is not the right place to learn how to use JUnit, but, seeing it in context can be very helpful and we already have the code right there.

@harsha509 what do you think?

rpallavisharma:
Sure.. fair enough .. (about test runners) 

 we may give pointers of different test runners available, and as you said showcase the ones we have examples with(but as a reader I would request we finalize from doc point of view which test runner for which language in our code examples to maintain harmony.. (couldn't think of another word..)

and also share them on guidelines, as in for contributors who would be submitting example codes on them in future. ) 

About how to name section.. we may name them as per the usage we are showcasing there.. 

-- i feel that the documentation.. is about the product which is selenium in here.. and then there is the usage aspect of it.. some segregation will be helpful.

titusfortner:
Ok, I've fixed things, renamed things and added an alert to show that it is more of a placeholder than complete content (all I have time to do right now).

https://deploy-preview-1479--selenium-dev.netlify.app/documentation/webdriver/getting_started/using_selenium/

I'm not sure we're actually going to end up seeing much work from the person we contracted with to help us because apparently she has gotten a new job, so we may have to do more ourselves. is this something we can work with? What do you think of this @pallavigitwork / @harsha509 ?

rpallavisharma:
Hi @titusfortner , I saw the page. Using Selenium (for xyz reasons , one of them testing) -- should be,  I think a different Section, as in Main, instead of in Getting Started. We need not talk about installing Selenium / getting started with Selenium again. 
I like the sectioning part you have created. but in the same i see inclusion of maven, gradle.. , talks about hooks / setting / tearing down, that considering someone who is trying to get started with Selenium may find difficult to grasp. 

-- it might be a good idea to step back and rethink about documentation from perspective of explaining product - Selenium, and -- if we want to talk about the usage-- then how do we plan and go towards it. 

rest I will DM you on slack.

titusfortner:
@pallavigitwork so you are suggesting we change this page (https://www.selenium.dev/documentation/overview/) to be more browser automation focused and less test automation focused? That makes sense to me.

So I should remove that section from — https://deploy-preview-1479--selenium-dev.netlify.app/documentation/webdriver/getting_started/using_selenium/ ??

> step back and rethink about documentation from perspective of explaining product

Yes, I think we all know we need this; this is what the PLC contracted a professional technical writer to help us with. I'm not sure yet the actual status.

rpallavisharma:
Hi @titusfortner yes you are right i am suggesting this "so you are suggesting we change this page (https://www.selenium.dev/documentation/overview/) to be more browser automation focused and less test automation focused? That makes sense to me." 

-- I think clarification from PLC regarding writing maybe required, to avoid double effort, and have more clarity. 

-- i have tried, to create a document of what I am thinking, using what you have already written, modifying where I find, its not complete but currently a thought process. 
[Selenium- Getting Started Page.docx](https://github.com/SeleniumHQ/seleniumhq.github.io/files/12645477/Selenium-.Getting.Started.Page.docx)

titusfortner:
This isn't perfect, but it's better than what we have, so I'm going to merge it. Thanks.

PR #1489, opened 2023-10-01
Revamp Bidi Documentation

titusfortner:
### Status
* Need to verify all the links
* Need to verify aliases
* Need to verify codeblocks
* Need to update translations

I really want to move the network interception code over, but I might have to wait, or ask other kind individuals to help out.
There is a lot more to do here, especially comparing some of these examples to the new WebDriver BiDi implementations.

### Description
* Reorganizes the section into 4 parts:
   - CDP with Endpoint examples
   - CDP API examples
   - CDP with BiDi API examples
   - WebDriver BiDi examples
 * Added a bunch code examples for many of these

### Motivation and Context
Current setup is a little confusing, and the link names are hard to parse
Many of the examples are mixing the different approaches in a way that didn't always make sense
A lot of the existing code is wrong. This way it all executes

I created duplicate examples in some of these to show the differences. It may be overkill, but I thought it useful to better understand why the examples online are all over the place.

### Review Help
These are the 4 pages I changed: https://deploy-preview-1489--selenium-dev.netlify.app/documentation/webdriver/bidirectional/chrome_devtools/
(I didn't change the WebDriver-BiDi content, just renamed the directory)

Do the page names make sense? Am I describing everything ok?


examples/dotnet/SeleniumDocs/Bidirectional/BidiApiTest.cs
```diff
@@ -72,5 +93,23 @@ public async Task InterceptNetworkRequest()
 
             StringAssert.Contains(driver.PageSource, "delicious cheese");
         }
+
+        private string IsDisplayedScript()
```
nvborisenko: I guess no private methods should be in examples. Any example should be self-readable. It's better to provide simple js code to be executed directly in example, rather than hide it from user as private.

examples/dotnet/SeleniumDocs/Bidirectional/BidiApiTest.cs
```diff
@@ -8,32 +11,50 @@ namespace SeleniumDocs.Bidirectional
...
 
+            var isVisibleDisplayed = ((WebDriver)driver).ExecuteScript(isDisplayed, visible);
+            var isHiddenDisplayed = ((WebDriver)driver).ExecuteScript(isDisplayed, hidden);
+
+            Assert.IsTrue((bool)isVisibleDisplayed);
+            Assert.IsFalse((bool)isHiddenDisplayed);
         }
 
         [TestMethod]
         public async Task InterceptNetworkResponse()
```
nvborisenko: image

this is how it looks for end user...

examples/dotnet/SeleniumDocs/Bidirectional/BidiApiTest.cs
titusfortner: This isn't going to be displayed in the docs page. Before hooks and private methods that are obviously named should be fine. We could do a trivial example that isn't as interesting, but this one is especially useful in the real world, so I prefer it.

examples/dotnet/SeleniumDocs/Bidirectional/BidiApiTest.cs
titusfortner: Thanks, I'll fix

pujagani:
Thank you so much for this. I like the write-up. It makes things clear and the examples help too. Overall, it looks good to me!

pujagani:
I verified most links, they work as expected.

website_and_docs/content/documentation/webdriver/bidirectional/chrome_devtools/cdp_endpoint.en.md
```diff
@@ -0,0 +1,102 @@
...
+title: "Chrome DevTools Protocol Endpoint"
+linkTitle: "CDP Endpoint"
+weight: 2
+description: >
+  Google provides a `/cdp/execute` endpoint that can be accessed directly. Each Selenium binding provides a method that
+  allows you to pass the CDP domain as a String, and the required parameters as a simple Map.
+---
+
+{{% pageinfo color="warning" %}}
+These methods will eventually be removed. It is recommended to use the [WebDriver-BiDi]({{< ref "../webdriver_bidi/" >}}) [WebDriver Bidi APIs]({{< ref "bidi_api.md" >}})
```
pujagani: Nit: Is there a conjunction missing between `[WebDriver-BiDi]({{< ref "../webdriver_bidi/" >}}) [WebDriver Bidi APIs]`? 
Screenshot 2023-10-04 at 5 56 12 PM

titusfortner: yup, will fix. thanks.

examples/dotnet/SeleniumDocs/Bidirectional/BidiApiTest.cs
titusfortner: Ok, I removed this after all to make it more simple

titusfortner:
Ok, overhauled everything to get all the tests passing in the CDP implementations and make comments in each part of the code for what users should actually use.

PR #1522, opened 2023-11-04
Remote WebDriver examples

titusfortner:
While I was adding code for downloads, I decided to move / add code for uploads (both local & remote) and examples for running browser specific features. 

PR to run the tests and double check the syntax


titusfortner:
That's a lot of failures...

PR #1529, opened 2023-11-17
Upgrade to Docsy 0 7 2

titusfortner:
### Description
The primary change was to tabpane code. The default is now to persist, so can remove all references to `langEqualsHeader=true` which never really made sense as the reason something got persisted. 

### Motivation and Context
Keep up to date with Docsy


titusfortner:
Moving it to draft because we still need `langEqualsHeader=true` when we are actually using the language as the header and I removed all of that.

titusfortner:
Ok, I fixed it. Should be good now.

diemol (approved):
Thank you, @titusfortner!

PR #1530, opened 2023-11-17
Tabpane updates

titusfortner:
### Description
Based on the new tab pane code (#1529 needs to be merged first)...

1. This allows us to set the active tab from the URL again:
https://selenium.dev/documentation/webdriver/actions_api/mouse/?tab=ruby#forward-click

2. This sets the tab height the same across each pane.

### Motivation and Context
* When linking to something in our docs, it would be nice to send users to the language directly
* It is really difficult to switch between tabs when the whole page resizes


titusfortner:
@diemol I think this is good to update.
I've PR'd both of these to Docsy, and they are looking into them for Q1: https://github.com/google/docsy/issues/1746
When they merge and release, we can update module version and delete this file.

PR #1532, opened 2023-11-24
Change how we do versioning

titusfortner:
### Description
* Remove the Legacy stuff from the sidebar and put it in a separate directory
* Add version toggle between "Documentation" and "Legacy" ~in the nav bar~ above breadcrumbs
* Change 4.0 to 4.15 as the "current version"
* Remove all the "This code requires Selenium xxx or greater" badges

### Motivation and Context
* ~drop down in nav bar is the default Docsy way to do it~ This was going to be the default Docsy way, but it didn't make sense to have versions always displayed in NavBar even when not on documentation page. Also, I customized the toggle so the dropdown gives the version being displayed, not a title hard-coded in the toml file
* The badges were apparently confusing
* The Legacy section really shouldn't be shown in the sidebar.

What do you think? Is this better?

### Update:
Because we are only translating the documentation, it might make sense to move that to the new place as well?

What do we think? Put it back?


titusfortner:
looks like this is on all pages, not just documentation page, so need to fix that....

diemol (commented):
This opens the door to maintain different versions of the documentation. Do we want that? Do we have the time and people to do that?

I am fine with removing the badges and doing something about it. But I am skeptical about having a version drop down.

titusfortner:
1. It's dumb to have Legacy stuff in the documentation, we at least should figure out someplace else to put that /somehow/
2. Right now we're putting badges on any code with a minimum version of what is needed for it to work. I actually like this, but I've been told it is confusing...
3. I do think we should be versioning our code, but it doesn't make sense to do this until we have all of our docs actually updated 😂

diemol:
> It's dumb to have Legacy stuff in the documentation, we at least should figure out someplace else to put that /somehow/

If `Legacy` is confusing, then why not hide it?

>To hide a page or section from the left navigation menu, set toc_hide: true in the front matter.

https://www.docsy.dev/docs/adding-content/navigation/#section-menu

titusfortner:
(in case it wasn't obvious, I'm the one who set Legacy to be where it is now, so I'm calling what I did previously dumb).
I'll take a look at hiding it.

PR #1534, opened 2023-11-25
Fix Mobile View

titusfortner:
I'm not sure why our navbar works differently than the default docsy one. I was looking at CloudWeGo because I thought it might be related to our dropdowns-inside-navbar, but in the end, removing `@extend .navbar-expand;` from the default docsy makes it all just work? (see last commit)

Maybe there are things that need to get "put back?" e.g.: https://github.com/cloudwego/cloudwego.github.io/blob/main/assets/scss/_nav.scss

But this looks much better than what we have right now in all sizes and emulations.

### DIFF
--- a/website_and_docs/assets/scss/_nav.scss
+++ b/website_and_docs/assets/scss/_nav.scss
@@ -1,0 +1,213 @@
+//
+// Main navbar
+//
+
+.td-navbar-cover {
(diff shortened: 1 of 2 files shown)


harsha509 (approved):
Awesome 🚀

diemol (approved):
Thank you, @titusfortner!

PR #1563, opened 2024-01-23
update example code to 4.17

titusfortner:
Running tests against 4.17


examples/java/src/test/java/dev/selenium/drivers/RemoteWebDriverTest.java
```diff
@@ -8,6 +8,7 @@
 import java.nio.file.Path;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Comparator;
```
titusfortner, line 11: reminder, this breaks the docs for everything on that page. Will fix then merge

PR #1586, opened 2024-02-22
Update first_script.py

kevinanewman:
Apparently submit_button.click() should be submit_button.submit()

### Types of changes

- [ ] Improved translation

### Checklist

### DIFF
--- a/examples/python/tests/getting_started/first_script.py
+++ b/examples/python/tests/getting_started/first_script.py
@@ -13,7 +13,7 @@ text_box = driver.find_element(by=By.NAME, value="my-text")
 submit_button = driver.find_element(by=By.CSS_SELECTOR, value="button")
 
 text_box.send_keys("Selenium")
-submit_button.click()
+submit_button.submit()
 
 message = driver.find_element(by=By.ID, value="message")
 text = message.text


mmonfared:
@kevinanewman This is the very first test. Although `submit()` works, I do believe we should include at least one `click()` here for demonstration purposes and to familiarize the reader with it.

kevinanewman:
Ok, `click()` by itself didn't seem to do anything, and didn't submit the form (unlike "clicking" the actual button with an actual mouse on the actual web page).  But `submit()` behaved as though clicking the button manually. 

Without a `submit()`, nothing happens, at least for me anyway.

mmonfared:
> Ok, `click()` by itself didn't seem to do anything, and didn't submit the form (unlike "clicking" the actual button with an actual mouse on the actual web page). But `submit()` behaved as though clicking the button manually.
> 
> Without a `submit()`, nothing happens, at least for me anyway.

Actually, it submits. The reason you might think nothing happened is because the driver quits immediately. If we pause before `driver.quit()`, you can see this confirmation page:

image

Also, I think it would be better to have an assertion here and keep this page visible for users by a simple sleep. So I would suggest adding these lines before `driver.quit()`:

```python
assert  text == "Received!"
sleep(1) # For demo purpose
```

diemol (commented):
These tests are executed daily and they are currently passing: https://github.com/SeleniumHQ/seleniumhq.github.io/actions/runs/7996701078. Why do you think it needs to be changed?

kevinanewman:
Well, that's interesting.  I wonder if maybe there should be a delay after editing the text box so you can see that it actually happened.  When I observed the `click()` apparently not working, I was executing the lines one at a time, not running the full script all at once.  When I run the full script using `webdriver.Chrome()` the script runs and the window disappears, when I use `webdriver.Safari()` the window stays open and shows `Received!`.  

So everything does appear to be working, although there are some subtleties apparently!  I can close this PR if you like.

diemol:
Yeah, I think we won't add this change as our CI passes.

kevinanewman:
Ok, thank you for your time and assistance!

PR #1607, opened 2024-03-06
doc: add more noob fdly code

AlphaWong:
# background
the quickstart for the javascript is oversimple. it takes time for noob like my self to dig into the source to check the rightway.

# file needs to be changed
https://github.com/SeleniumHQ/seleniumhq.github.io/blob/trunk/examples/javascript/test/getting_started/firstScript.spec.js

# propose
```js
// doc: https://www.npmjs.com/package/selenium-webdriver?activeTab=code
const {Builder, Browser} = require('selenium-webdriver');

(async function helloSelenium() {
  let driver = await new Builder().forBrowser(Browser.CHROME).build();

  await driver.get('https://selenium.dev');

  await driver.quit();
})();
```
I would create a PR for update the doc for noob to pickup the framework more smoothly

### DIFF
--- a/examples/javascript/test/getting_started/firstScript.spec.js
+++ b/examples/javascript/test/getting_started/firstScript.spec.js
@@ -1,4 +1,3 @@
-// doc: https://www.npmjs.com/package/selenium-webdriver?activeTab=code
 const {By, Builder, Browser} = require('selenium-webdriver');
 const assert = require("assert");
 
--- a/examples/javascript/test/getting_started/firstScript.spec.js
+++ b/examples/javascript/test/getting_started/firstScript.spec.js
@@ -1,11 +1,12 @@
-const {By, Builder} = require('selenium-webdriver');
+// doc: https://www.npmjs.com/package/selenium-webdriver?activeTab=code
+const {By, Builder, Browser} = require('selenium-webdriver');
 const assert = require("assert");
 
 (async function firstTest() {
   let driver;
   
   try {
-    driver = await new Builder().forBrowser('chrome').build();
+    driver = await new Builder().forBrowser(Browser.CHROME).build();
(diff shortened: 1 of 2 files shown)


diemol:
How is `Browser.CHROME` making it more simple?

examples/javascript/test/getting_started/firstScript.spec.js
```diff
@@ -1,11 +1,12 @@
-const {By, Builder} = require('selenium-webdriver');
+// doc: https://www.npmjs.com/package/selenium-webdriver?activeTab=code
```
diemol: If you add one more line, the references get broken, why is this line needed?

AlphaWong:
> How is `Browser.CHROME` making it more simple?

as I need to check this file to find the source of Enum
https://github.com/SeleniumHQ/selenium/blob/trunk/javascript/node/selenium-webdriver/lib/capabilities.js#L31

if I know there is a Emun list on the quickstart page, a noob like myself can easily know the expected input rather than looking at the source and getting confused for the README at https://github.com/SeleniumHQ/selenium/tree/trunk/javascript/node/selenium-webdriver#usage

```js
/**
 * Recognized browser names.
 * @enum {string}
 */
const Browser = {
  CHROME: 'chrome',
  EDGE: 'MicrosoftEdge',
  FIREFOX: 'firefox',
  INTERNET_EXPLORER: 'internet explorer',
  SAFARI: 'safari',
}
```

examples/javascript/test/getting_started/firstScript.spec.js
AlphaWong: good point, i should remove this line

AlphaWong: deleted! thx for review!

diemol (approved):
Thank you, @AlphaWong!

PR #2625, opened 2026-04-21
Fix preStop behavior for KEDA scale-down

kool79:
### Description
Updated the preStop hook in example to decouple it from Selenium node command-line arguments. This prevents random node terminations during KEDA scale-down when use the standard docker images for nodes.
Removed "s" from the 'sleep' command to ensure compatibility with POSIX

### Motivation and Context
The previous implementation of the preStop hook relied on Selenium node command-line arguments to delay shutdown until session finished. This made the logic fragile and dependent on how the node process was started, which could vary between environments or deployments. 
_In practice, this approach failed for the standard Docker Chrome node container, where the expected command-line structure is different._
As a result, during KEDA scale-down events, nodes were sometimes terminated unexpectedly before properly handling ongoing sessions.

### Types of changes

- [ ] Improved translation

### Checklist


rpallavisharma:
@VietND96  / @diemol please see the PR for review

diemol (approved):

PR #2639, opened 2026-05-14
Add Python examples for W3C BiDi network intercept commands

### LINKED ISSUE #2638: [🚀 Feature]: Add Python examples for W3C BiDi network intercept commands
## Problem
The W3C BiDi Network documentation page currently contains 
Java examples only for network intercept commands (add intercept, 
remove intercept, fail request, request handlers).

Python examples are missing entirely.

## Proposed Solution
Add a new file `examples/python/tests/bidi/test_network_commands.py` 
with Python equivalents mirroring the existing Java tests and 
following the pattern established in `test_bidi_logging.py`.

## Scope
- test_add_intercept
- test_remove_intercept  
- test_fail_request
- test_add_and_remove_request_handler

I am happy to submit a PR for this if the team agrees this is needed.

ajithrao2509:
Closes #2638

### Description
Adds Python test examples for the W3C BiDi Network intercept commands, and mirrors the Python tabs across all translated docs.

Files changed:
- `examples/python/tests/bidi/test_network_commands.py` — new Python test file with 4 tests
- `website_and_docs/content/documentation/webdriver/bidi/w3c/network.en.md` — Python tab added to each section
- `website_and_docs/content/documentation/webdriver/bidi/w3c/network.ja.md` — Python tabs mirrored
- `website_and_docs/content/documentation/webdriver/bidi/w3c/network.pt-br.md` — Python tabs mirrored
- `website_and_docs/content/documentation/webdriver/bidi/w3c/network.zh-cn.md` — Python tabs mirrored

### Motivation and Context
The W3C BiDi Network documentation page had Java and JavaScript examples only. Python examples were completely absent — not even a badge-code placeholder. This PR adds Python equivalents for the commands we have tests for, honest badge-code placeholders for the ones we don't, and updates all translated language pages per the contributing guide.

### Types of changes
- [x] Code example added
- [x] Code example added to all translated languages
- [ ] Change to the site
- [ ] Improved translation

### Checklist
- [x] Follows existing file and naming conventions
- [x] Each test has an assertion
- [x] License header included
- [x] `@pytest.mark.driver_type("bidi")` marker on every test
- [x] `gh-codeblock` references use verified line numbers
- [x] `badge-code` placeholder used honestly where no test exists
- [x] Issue #2638 raised before this PR
- [x] Review feedback addressed (observable assertions, TimeoutException waits, safe intercept cleanup, trailing newline)


rpallavisharma:
please add details for all languages. please see the contributing doc - https://www.selenium.dev/documentation/about/contributing/

ajithrao2509:
> please add details for all languages. please see the contributing doc - https://www.selenium.dev/documentation/about/contributing/

Hi @rpallavisharma the details have been added to all the languages according to the principles from the contribution documentation.

diemol (dismissed):
The new test seems to be getting stuck; the workflow does not end in a timely manner. The one in `trunk` does.

Claude tells me this:
>In test_fail_request, driver.network.fail_request(request.request["requestId"]) is wrong — Network has no fail_request method. fail_request() actually lives on the request object passed into the callback, so it
  should just be request.fail_request().  

Please double-check the test, and we can run CI again.

ajithrao2509:
Following up on the driver.get() hang in test_add_and_remove_request_handler that CI flagged.

**What I found:**

Reproduced it locally, consistently not a  CI-runner flakiness.
Isolated the test on its own (pytest ... ::test_add_and_remove_request_handler) — still hangs, so it's not caused by test ordering or leftover state from other tests.

The original code never called request.continue_request() inside the before_request callback. Per BiDi's design, registering a before_request handler blocks the browser's navigation until the callback explicitly continues it — so driver.get() was waiting indefinitely for a response that was never sent, until the 120s HTTP client timeout hit.

Adding request.continue_request() to the callback fixed the hang, but surfaced a new, intermittent KeyError: <id> inside Selenium's own websocket_connection.py, at self._messages.pop(current_id). Tracing it: continue_request() runs on the BiDi event-callback thread, while the main test thread is simultaneously issuing WebDriver commands over the same connection both share one self._id counter and self._messages dict with no locking, so the two threads can race on the same message ID.

**What I did:**
Given this looks like a thread-safety issue in the Python BiDi client itself rather than something fixable from the test, I've marked this one test as skipped with the full reasoning inline, rather than continuing to patch around it. The other 3 tests in the file pass cleanly and quickly now (3 passed, 1 skipped, no more 5+ minute hangs).

Happy to open a separate issue against the main selenium repo for this if that'd be useful. let me know, or if there's a known-safe pattern here I'm missing @diemol

diemol:
It seems that this is still failing. Have you tried this locally? Is it passing for you?

ajithrao2509:
> It seems that this is still failing. Have you tried this locally? Is it passing for you?

Checked both nightly failures, neither is in my changed files. tests/browsers/test_console_log.py fails on Windows nightly with a file-locking PermissionError, and tests/actions/api/test_mouse.py::test_move_by_offset_from_element fails on Ubuntu nightly with a ValueError parsing coordinates. Both look like pre-existing issues on the nightly channel, unrelated to this PR. I did also notice a warning (not a failure) in test_fail_request 'Request' object has no attribute 'fail_request' on nightly which suggests possible API drift between the stable and nightly Selenium builds.

diemol:
No, I don't mean the other failures. I meant the specific test you are adding.

ajithrao2509:
> No, I don't mean the other failures. I meant the specific test you are adding.

To directly answer: no, it did not pass for me locally, consistently I ran it in isolation multiple times and it fails the same way each time with the KeyError as I described. Happy to share the exact local run output if useful. Given that, is the skip the right call here, or would you prefer I keep digging into a fix on my end first?

ajithrao2509:
hi @diemol Pushed an update that removes the skip and fixes the underlying issues. Summary of what I found and changed:

**Root cause of the `KeyError`:** an unsynchronized message-ID race in `WebSocketConnection.execute()` in selenium 4.41.0 (which `examples/python/requirements.txt` pins) BiDi event callbacks run on separate threads, and concurrent `execute()` calls can collide on the shared `_id`/`_messages` state. This was fixed upstream in selenium 4.44.0 via `_id_lock`. Rather than keep the test skipped until the pin is bumped, `test_add_and_remove_request_handler` now follows the same structure as the same-named test in `py/test/selenium/webdriver/common/bidi_network_tests.py`: the handler is removed before navigating, so no request is ever blocked awaiting continuation which avoids both the hang and the race on the pinned version.

**While validating locally, I also found `test_fail_request` was passing without exercising the command.** With classic `driver.get()`, `network.failRequest` consistently returns `Invalid InterceptionId` on Chrome, and the expected `TimeoutException` occurred only because the blocked request was never actually failed. This is consistent with the classic-navigation skip in the selenium test suite ("Request handlers don't yet work in Chrome using classic navigation"). The test now navigates via `browsing_context.navigate` the same approach the suite's fail-request tests use and asserts the raised `WebDriverException`, so the example genuinely demonstrates the command. The callback guards `fail_request()` with a narrow `except WebDriverException`, because the event subscription also delivers events for requests already cancelled by the aborted navigation. I also removed a redundant manual intercept (`add_request_handler` already creates one internally).

All four tests pass locally on selenium 4.41.0 + Chrome, no skips, no warnings:

```
tests/bidi/test_network_commands.py::test_add_intercept PASSED
tests/bidi/test_network_commands.py::test_remove_intercept PASSED
tests/bidi/test_network_commands.py::test_fail_request PASSED
tests/bidi/test_network_commands.py::test_add_and_remove_request_handler PASSED
==================================================== 4 passed in 16.05s ==========================================
```
Docs line ranges updated across all four language pages to match the final file.

**On the expected CI results:** stable should be green on both OSes. The two nightly failures visible on recent runs  `test_move_by_offset_from_element` (ValueError, Ubuntu) and `test_log_level` (PermissionError, Windows) are pre-existing on the nightly channel and not in this PR's files; the scheduled nightly runs on `trunk` are currently failing as well, independent of any PR. The only expected reference to this PR's file on nightly is a warning from the `fail_request` → `fail()` API rename on trunk.

**Note for later:** selenium 4.44+ replaced this API with a high-level handler form (`request.fail()`, automatic continuation). When the examples' selenium pin is next bumped, I'd be happy to modernize these examples in a follow-up PR mirroring the current `bidi_network_tests.py` patterns.

diemol:
So, what was the issue? I tend to avoid reading a copy & pasted text from AI.

ajithrao2509:
> So, what was the issue? I tend to avoid reading a copy & pasted text from AI.

The actual was the examples pin selenium 4.41, which has a threading bug in the BiDi websocket client (no lock on command ids, fixed in selenium 4.44). My callback was sending a command while another event was being processed and the responses collided, that was the KeyError. I rewrote the test the same way selenium's own test_add_and_remove_request_handler works (remove the handler before navigating) so nothing is blocked and the bug never triggers, skip is gone. While retesting I also found test_fail_request was passing without actually failing the request, Chrome rejects failRequest during driver.get, so it now uses browsing_context.navigate like selenium's own tests. All 4 pass locally on 4.41 + Chrome.

diemol:
Interesting. Now we bumped the test suite to 4.46, and this new test is failing again, could you please double check?

https://github.com/SeleniumHQ/seleniumhq.github.io/actions/runs/29239632949/job/86782316021?pr=2639#step:11:63

ajithrao2509:
> Interesting. Now we bumped the test suite to 4.46, and this new test is failing again, could you please double check?
> 
> https://github.com/SeleniumHQ/seleniumhq.github.io/actions/runs/29239632949/job/86782316021?pr=2639#step:11:63

hi @diemol I have made the changes as necessary, earlier when I ran them in my local the selenium version was 4.41, so my validation was against the old pin. On 4.46 the request object has fail() instead of fail_request() which is why the tests broke after the version bump. I rewrote the test_fail_request using the 4.46 handler API same as test_fail_requests_matching_url_pattern in the selenium test suite and verfied the tests locally which are currently passing without any failures on Chrome. Updated the document line ranges as well.

diemol (commented):
Great, thank you for your effort!

diemol (approved):

